PR #45, opened 2014-02-14
Keep ship_method_ref as a simple string

huoxito:
Even though quickbooks api tells it is a ReferenceType I don't any
ShipMethod type on their docs or UI. Setting a name to the shipping
method works just fine though. The name is properly displayed on the
sales receipt page.

Does that make sense?

Sales order reference doc: https://developer.intuit.com/docs/0025_quickbooksapi/0050_data_services/030_entity_services_reference/salesreceipt

### DIFF
--- a/lib/quickbooks/model/sales_receipt.rb
+++ b/lib/quickbooks/model/sales_receipt.rb
@@ -19,7 +19,7 @@ module Quickbooks
 
       xml_accessor :po_number, :from => 'PONumber'
 
-      xml_accessor :ship_method_ref, :from => 'ShipMethodRef', :as => BaseReference
+      xml_accessor :ship_method_ref, :from => 'ShipMethodRef'
       xml_accessor :ship_date, :from => 'ShipDate', :as => Time
       xml_accessor :tracking_num, :from => 'TrackingNum'
 


HeathNaylor:
@huoxito @ruckus this should be a BaseReference, just setting as a string does not work.

ruckus:
@HeathNaylor  you're absolutely right. I realize that I went too fast with this PR and I understood it as being the reverse: the attribute lacked the `BaseReference` casting and it was previously just a `String` - basically the opposite of what the PR. That would have been a good thing. 

Alas, its a regression, as that attribute should be a `BaseReference`. I'm going to reverse this commit.

ruckus:
Argh, maybe not. The docs at:

https://developer.intuit.com/docs/api/accounting#/salesreceipt

Have this to say about **ShipMethodRef**: 

> Reference to the ShipMethod associated with the transaction. There is no shipping method list. Reference resolves to a string.

So, now doing *another* reversal, it seems treating it as a String is correct.

huoxito:
hey guys, it's been over a year I dont touch this code but yeah I clearly remember this working as a string, perhaps there's some per account configuration we're missing

HeathNaylor:
I just attempted a string and had exceptions thrown, when I passed in as a reference it works fine. Not sure the API is accurate in what is accepted here.

ruckus:
@HeathNaylor  where did you get a Ship Method ID to use in the reference setter? I dont see an endpoint for *ShipMethod* to fetch a list of them..?

HeathNaylor:
This is all I did.

```
shipping_reference = Quickbooks::Model::BaseReference.new
shipping_reference.name = 'String Here'
shipping_reference.value = 'String Here'
```
This would show "String Here" on the QBO invoice.

The signature for the ShipMethodRef in the QBO API doesn't need an ID.
```
name: string 
type:	string 
value: string, required 
```

ruckus:
Gotcha, thanks. Ok, I will re-insert the base reference and add something to the docs about this.

HeathNaylor:
Thanks, sorry for the confusion, I honestly don't see why they don't just allow string, but we can only do what we can do I suppose.

PR #205, opened 2015-01-04
Namespace Validator class

bradleypriest:
It's only used in Quickbooks::Model::BaseModel, so the including should still "just work"


ruckus:
Thank you!

PR #207, opened 2015-01-13
Add ability to fetch PurchaseTaxRateList from TaxCode

gherry:
Parse list of TaxRate that is necessary for Bill and/or PurchaseOrder's TaxLine


ruckus:
Thank you!

PR #210, opened 2015-01-19
Add HasLineItem module to DRY transaction initialization

gherry:
I found out that `Bill` and `JournalEntry` do not ensure line items to be empty array. Instead of copying the `initialize` method in all these transactions classes, the line items specific logic is moved to `HasLineItems` module to reduce repeated logic


ruckus:
Thank you!

PR #226, opened 2015-03-08
Add Deposit Element

ericgascoine:
I've created and added the Deposit type as per  https://developer.intuit.com/docs/api/accounting#/deposit-beta as I'm needing to create deposits that have payment charges offsetting the actual payments.

I'm aware that this may need to be updated once this is fully finalized with Quickbooks.

This is currently tested and working with the sandbox environments


minimul:
Nice Eric. I went thorough the PR, looks good.

ruckus:
Thank you!

ericgascoine:
You're welcome, but I'm discovering that the test coverage was insufficient when I'm building the integration components for  our/my project.

I'm in the process of some refinements and expansion of the test cases to make this addition a little more robust, and hope to have multiple forms of deposits running within a couple of weeks (This is a side project rather than paying work).  So far the query and reading is working completely, but the injection is running into a few issues which I'm diagnosing.

I'll keep you informed, if something comes up in the short term, feel free to ping me on this if you need to before hand.

I also noticed the removal of the puts statement.... Sorry about leaving that in there.

PR #234, opened 2015-03-31
Update README.md to indicate that `line_item` arrays need initialized

hrrf:
Initialize the `line_items` arrays for the given Models.


minimul:
I grep'd the lib/quickbooks/models directory. There are no models that have line_items but DON'T have the HasLineItems. This can be closed.

PR #235, opened 2015-04-01
Purchase requires a line item as a minimal requirement.

minimul:
Add a basic service create spec as well.


ruckus:
Thank you! I also added `include HasLineItems` to some other models which have `:line_items`: f2531bcc5ea6768e2ab7772c7fc2a7f7d90b263e

minimul:
Perfect. I was just going to do that.

PR #239, opened 2015-04-08
For attachable entity reference use base reference instead of entity reference

minimul:
Reference: https://github.com/ruckus/quickbooks-ruby/issues/238

attachable-entity-xml

I have a real test against the sandbox here: https://github.com/minimul/minimulcasts/blob/spec/spec/requests/qbo_spec.rb#L51

For the gem I thought it was good enough to just spec the request xml.


ruckus:
Thanks @minimul !!

PR #240, opened 2015-04-09
Check that reconnect_token_at is older than the current date instead of ...

HeathNaylor:
Check that reconnect_token_at is older than the current date instead of newer.


ruckus:
Thank you!

PR #247, opened 2015-04-29
Reports with options

Craggar:
Added 'options' to reports, allowing for more complex report requests than just report name + macro - e.g. specific start_date and end_date, configuring columns, accounting method, etc. Included a test for an example report with options


ruckus:
Thank you!

PR #249, opened 2015-05-01
 Assign EntityRef for Journal Entry to use BaseReference so that 'type' a...

minimul:
...nd 'name' attributes can be supported

e.g.
```
<?xml version="1.0"?>
<JournalEntry>
  <Line>
    <JournalEntryLineDetail>
      <Entity>
        <Type>Customer</Type>
        <EntityRef name="James Rockenstall">1</EntityRef>
      </Entity>
    </JournalEntryLineDetail>
  </Line>
</JournalEntry>
```


ruckus:
Thank you!

PR #250, opened 2015-06-04
Refactor change data capture related services/models. Add change data

michaelcheung:
capture services/models for payments and credit memos.


minimul:
Thanks, looks good.

arthurchui:
+1

Issue #253, opened 2015-06-29
BaseReference initialize change


divya-eligible:
Can the BaseReference(https://github.com/ruckus/quickbooks-ruby/blob/master/lib/quickbooks/model/base_reference.rb) initialize be changed to take type, name and value? I accidentally gave entity type, name and value during initialization and it wasn't working. later checked the initialize and figured out that the initialize takes only value.

ruckus:
Sounds like that could be useful. Care to create a Pull Request?

divyak-eligibleapi:
Sure.

PR #254, opened 2015-06-30
Changed the initialize method of BaseReference model

divyak-eligibleapi:
Related to https://github.com/ruckus/quickbooks-ruby/issues/253

Divyas-MacBook-Pro:quickbooks-ruby divyakesapragada$ git push origin master
Counting objects: 20, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (20/20), done.
Writing objects: 100% (20/20), 3.18 KiB | 0 bytes/s, done.
Total 20 (delta 14), reused 0 (delta 0)
To git@github.com:divyak-eligibleapi/quickbooks-ruby.git
   6c46373..dc2dedc  master -> master
Divyas-MacBook-Pro:quickbooks-ruby divyakesapragada$ bundle exec rspec spec
..................................................................................................................................................................................................................................................................................................................................................................

Finished in 1.73 seconds
354 examples, 0 failures
Coverage report generated for RSpec to /workspace/quickbooks-ruby/coverage. 2518 / 2634 LOC (95.6%) covered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ I say *pseudo* because they look like Integers but Intuit has made it clear they
 Anyways, its subtle but the value for a `SalesReceipt#ShipMethodRef` while it is a `BaseReference` needs to be set manually:
 
 ```ruby
-shipping_reference = Quickbooks::Model::BaseReference.new(name: 'FedEx', value: 'FedEx')
+shipping_reference = Quickbooks::Model::BaseReference.new('FedEx', name: 'FedEx')
 receipt.ship_method_ref = shipping_reference
 ```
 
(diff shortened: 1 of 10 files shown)


minimul:
I like this PR but it appears to break `Quickbooks::Model::BaseReference.new(1)`.

We are going to have to support that "old" way of initializing BaseReference.

divyak-eligibleapi:
@minimul added backwards compatibility

ruckus:
Rather than make the whole argument a hash I think it makes more sense to make the `Id` parameter required and the other fields optional since that more closely resembles reality: the Id is required and the bare minimum and the name/type parameters are extra meta-data.

So the signature might look like:

`def initialize(id, options = {})`

and invocation would look like:

`Quickbooks::Model::BaseReference.new(1, name: "FedEx", value: "FexEx")`

Thoughts?

minimul:
@ruckus Yup, well said. 

@divyak-eligibleapi Would you implement `def initialize(id, options = {})` ?

divyak-eligibleapi:
sure makes sense.

minimul:
Specs failing.

divyak-eligibleapi:
yes i will check and update. got busy with some work. will update it today.

divyak-eligibleapi:
@ruckus @minimul fixed the issue that was causing specs failure. can you please check now?

ruckus:
Great, thank you!

PR #255, opened 2015-07-09
Move Definition module inside of Quickbooks::Model module.

ratbeard:
This fixes a bug in my rails application which contained a model
named Definition.  In staging when models were eager loaded, an
error is thrown because this gem defines a top level module called
Definition first.  This module should be nested inside the gems
module namespace and not be at the top level to avoid this clash.


ruckus:
Makes total sense - thank you!

Issue #146, opened 2014-07-31
Feature: Implement Reports API


cdmwebs:
It looks like QBO recently exposed a [reports API](https://developer.intuit.com/docs/0025_quickbooksapi/0050_data_services/reports). I'd like to take a stab at implementing some of the more important reports (Balance Sheet, P&L). 

Any tips as to approach? I was planning to create a Report base class and inherit for the individual reports, but I'm not sure how much is shared.

ratbeard:
I am wondering if anyone has gotten the code in #204 that was merged in to work.  I always get an empty Collection response when I try to use the ReportService.

Looking at the code, ReportService tries to parse the response xml as an collection, which expects the xml response root to be an <IntuitResponse>[1].  As I query the API, and look at the examples inside the spec/fixtures directory, I'm seeing the root node is <Report> [2].  Therefore this `if` branch [3] never gets run, and an empty Collection is always returned.  The tests pass because they only checking the fixture xml contains a certain node, not that the a proper Report model is returned [4].

Maybe Intuits api changed, or as mentioned in the discussion in #204 maybe people are doing something like:

```rb
client.query('BalanceSheet')
xml = Nokogiri.parse(client.last_reponse_xml)
...
```

If thats the only way to get access to the underlying report data, then I don't understand the purpose of Models::Report or code like https://github.com/ruckus/quickbooks-ruby/blob/e3b2cb50ff4605e625c0dffd54b752a753bfa23c/lib/quickbooks/service/base_service.rb#L125-L127 as they are never run.

I plan to work on this so at the minimum client.query returns a Report object with a `body` attribute containing the xml, and just wanted to check in to make sure I'm not missing something; perhaps a way to call Intuits report api and get back a collection of reports?  Otherwise I'll just rip out some of the code that got added and won't worry too much about breaking the api as it currently does not work.

Thanks!
:rat:

[1] https://github.com/ruckus/quickbooks-ruby/blob/e3b2cb50ff4605e625c0dffd54b752a753bfa23c/lib/quickbooks/service/reports.rb#L22
[2] https://github.com/ruckus/quickbooks-ruby/blob/e3b2cb50ff4605e625c0dffd54b752a753bfa23c/spec/fixtures/balancesheet.xml
[3] https://github.com/ruckus/quickbooks-ruby/blob/e3b2cb50ff4605e625c0dffd54b752a753bfa23c/lib/quickbooks/service/base_service.rb#L101-L102
[...]

Issue #257, opened 2015-07-22
Is there any plans to add Tax API supports?


raecoo:
Currently, only the Tax Code query api available. Actually, we also need other Tax APIs. e.g. create and update.

minimul:
Everything is done on a volunteer and "as needed" (usually) basis. Since you need these abilities it would be great if you could add them back into the gem. By looking at the specs in combination with https://developer.intuit.com/docs/api/accounting it is easy to get started implementing these features.

raecoo:
@minimul I tried to implement the TaxService API, but encountered the exception "System Failure Error: Cannot consume content type", Can you help to take a look?

this is my fork version https://github.com/raecoo/quickbooks-ruby
here is exception info https://gist.github.com/raecoo/b237c267127837867686
the request params and format is correct as official API docs mentioned.

thanks.

minimul:
Where is service/tax_service.rb?

raecoo:
@minimul Sorry, forgot committed. please pull it again.

tchang1:
@raecoo Tax service is JSON only. https://developer.intuit.com/v2/apiexplorer?apiname=V3QBO#?id=TaxService

raecoo:
@tchang1 Yes, I have already implemented the feature in my fork version, will push it later.

raecoo:
@minimul I've pushed a beta version to my fork that implemented the TaxService API via JSON way.

minimul:
I know I was just looking at it. Good find and work but at the same time Intuit should be consistent and provide the same support for both data formats. 

For better or worse this gem is wedded to XML requests and responses so I would like to see them flip the switch on XML for this entity. 

I just put in a question https://intuitdeveloper.lc.intuit.com/questions/1195446-taxservice-is-json-only . We need to evaluate their response before we can consider a merge.

raecoo:
Yup, I agree with you about the gem is wedded to XML requests and responses. so let's waiting their feedback then consider next thing.

minimul:
Update: There was some more comments on this thread: https://intuitdeveloper.lc.intuit.com/questions/1195446-taxservice-is-json-only

tchang1:
By the way, I'm from Intuit (you may have seen me answer q's on the intuit live community) but I lurk here to see if theres any issues I can help with. The short answer is we would prefer clients to use JSON and will not be enhancing existing services for xml support.

[...]

PR #259, opened 2015-07-27
Reports

ratbeard:
Related to discussion in #146.  Changes are:

- ReportsService.query now returns an instance of a Report model, instead of an empty intuit collection.
- Reports model renamed to Report, as its always a single item.
- Adds Report#xml method which returns the nokogiri doc,  similar to how the api was used before via `service.last_response_xml`
- Adds Report#columns method which returns an array of column names as strings, e.g. `report_with_years.columns.should == ["", "Jul 4 - Dec 31, 2013", "Jan - Dec 2014", "Jan 1 - Jun 21, 2015"]`
- Adds Report#all_rows method which returns an array of all <Row> or <Summary> in the document, e.g. things that contain <ColData>.  Nesting in Rows is ignored.  A row is simple to work with, looking like: `["Total Expenses", BigDecimal(100), BigDecimal(80)]`
- Adds Report#find_row(label) which is a simple helper to find a row by its label, e.g. `report.find_row("Total Expenses")`
- Add and cleanup some specs, including a new fixture data for a report with multiple columns.

~Thanks! :bow:


lib/quickbooks/model/reports.rb
```diff
@@ -1,17 +0,0 @@
-module Quickbooks
-  module Model
-    class Reports < BaseModel
-      XML_COLLECTION_NODE = "Reports"
-      XML_NODE = "Reports"
-      REST_RESOURCE = 'reports'
-
-      REPORT_TYPES = ['BalanceSheet', 'ProfitandLoss', 'ProfitAndLossDetail', 'TrialBalance', 'CashFlow', 'InventoryValuationSummary', 'CustomerSales', 'ItemSales', 'DepartmentSales', 'ClassSales', 'CustomerIncome', 'CustomerBalance', 'CustomerBalanceDetail', 'AgedReceivables', 'AgedReceivableDetail', 'VendorBalance', 'VendorBalanceDetail', 'AgedPayables', 'AgedPayableDetail', 'VendorExpenses', 'GeneralLedgerDetail']
```
ratbeard: These weren't used and were out of date w/ all the supported reports now, so I removed them

spec/lib/quickbooks/service/reports_spec.rb
```diff
@@ -1,68 +1,51 @@
...
+      url = @service.url_for_query('BalanceSheet', 'Last Year',
+        :start_date => '2015-01-01',
+        :end_date => '2015-01-31',
+        :accounting_method => 'Cash',
+        :columns => 'subt_nat_amount,tax_amount',
+      )
+      url.should == 'https://quickbooks.api.intuit.com/v3/company/9991111222/reports/BalanceSheet?start_date=2015-01-01&end_date=2015-01-31&accounting_method=Cash&columns=subt_nat_amount,tax_amount'
+    end
+
+    it 'currently ignores the date_macro argument when passed in additional options' do
```
ratbeard: This current behavior seems like a bug to me.  It seems like date_macro should just be another param in the options hash, and not be a positional argument.

minimul:
Thanks. Looks good to me, although, I am not using the existing reporting implementation in production so.. 
@Craggar , @raksonibs  any feedback?

minimul:
Hang tight on this. This breaks the previous API so we need to evaluate.

ratbeard:
The api has changed, but as people commented in the issue thread people aren't actually using the return value of `service.query`, they just use `service.last_response_xml` to get at the data.  The reason people don't use the return value is the code assumes the response comes back with a root <IntuitResponse> tag, which it does not, and therefore it always parses it as an empty collection.  Due to this, the Report model was never instantiated by the library either.

So basically the old way the api was used - `service.last_response_xml` - is still there, but now `query` also returns a meaningful Report model, which has a few additional helper methods for common cases of pulling out data.

No rush to merge and please confirm the above is true, I was a little confused by its behavior before and was wondering how the Report model ever worked.  Looking at the git history, it looked like there was a bad fixture at one point which contained several <Report>'s wrapped in an <IntuitResponse> which would have made the code work, but I haven't seen an actual response come back from quickbooks like this.  Hmm I wonder if the old code might work if you can call reports through a bulk api call?

minimul:
Ok, I am going to take this PR for a spin early next week.

Craggar:
Sorry, I've been away on vacation so just saw this now - I'll take a look at some point next week, but I'd happily accept having to make a couple of changes to use this, which actually returns an object. Looks good, at a cursory glance.

minimul:
Thanks @Craggar. I said I was going to take this for a spin but instead I am going to help out with https://github.com/ruckus/quickbooks-ruby/issues/257

@ruckus If @Cragger confirms why don't we merge and push as 0.3.0 after pushing what is in current master as 0.2.4

ruckus:
This looks great - thank you @ratbeard 

It sounds like I should merge this and release a `0.3.0` - since this introduces backwards incompatibility hence the minor version bump.

Let me know if y'all are ready. Thanks again.

minimul:
@ruckus It may be best if you `gem push` the current master at `0.3.0` and then when this PR is merged go to `0.4.0`. The recent JSON merge doesn't break backwards compatibility but was large enough to warrant a bump.

minimul:
Update: Would like to get this evaluated and potentially merged soon.

ratbeard:
Let me know if I can do anything to help.

I considered making a Row class to hold values for each row but ended up just using an array e.g: `['Savings', BigDecimal("19.99"), BigDecimal("19.99"), BigDecimal("819.99")]`

Craggar:
Hey, sorry I've not commented back for a while - I gave this a whirl, and it looks good to me. Happy to see this merged in.

minimul:
Thanks @Craggar 
We'll merge this in next week and bump the version to 0.4.0

minimul:
Thanks @ratbeard & @Craggar – merged.

PR #264, opened 2015-08-27
Introduce ActiveRecord like methods to Quickbooks Models

kashif-umair:
- This PR introduces some methods i.e. `all, where, find_by, find, count` methods to quickbooks models. This change eliminates the requirement of using services to query something from Quickbooks. You can directly call `Quickbooks::Models::Invoice.all` to get all the invoices.
- ~~Another setting `allow_pagination` is introduced in this PR. This is a boolean setting which is set to `true` by default. If user sets this setting to `false` then no pagination will be applied and all the records will be returned on any query.~~

---
- The golbal setting `allow_pagination` has been removed which I introduced in my last commit. Instead of that, I've introduced another option to skip pagination on each request basis i.e. `skip_pagination`. This setting will be available in both the conventional query method and the new methods introduced in this PR.

- [ ] I'm working on updating the `README.md` file.

### DIFF
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -17,6 +17,7 @@ require 'quickbooks/util/query_builder'
 #== Models
 require 'quickbooks/model/definition'
 require 'quickbooks/model/validator'
+require 'quickbooks/model/active_record_scaffold'
 require 'quickbooks/model/base_model'
 require 'quickbooks/model/base_model_json'
 require 'quickbooks/model/base_reference'
@@ -162,21 +163,12 @@ require 'quickbooks/service/payment_change'
 
 module Quickbooks
   @@sandbox_mode     = false
-  @@allow_pagination = true
 
   @@logger = nil
 
   class << self
     def sandbox_mode
       @@sandbox_mode
-    end
-
-    def allow_pagination?
-      @@allow_pagination
-    end
-
-    def allow_pagination=(allow_pagination)
-      @@allow_pagination = allow_pagination
     end
 
     def sandbox_mode=(sandbox_mode)
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -13,6 +13,7 @@ require 'quickbooks/util/logging'
 require 'quickbooks/util/http_encoding_helper'
 require 'quickbooks/util/name_entity'
 require 'quickbooks/util/query_builder'
+require 'configuration'
 
 #== Models
 require 'quickbooks/model/definition'
@@ -217,14 +218,6 @@ module Quickbooks
     def initialize(msg)
       self.message = msg
       super(msg)
-    end
-  end
-
-  class Configuration
-    def self.set_tokens_and_realm_id token, secret, realm_id
-      Thread.current[:token]    = token
-      Thread.current[:secret]   = secret
-      Thread.current[:realm_id] = realm_id
     end
   end
 
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -161,13 +161,22 @@ require 'quickbooks/service/credit_memo_change'
 require 'quickbooks/service/payment_change'
 
 module Quickbooks
-  @@sandbox_mode = false
+  @@sandbox_mode     = false
+  @@allow_pagination = true
 
   @@logger = nil
 
   class << self
     def sandbox_mode
       @@sandbox_mode
+    end
+
+    def allow_pagination?
+      @@allow_pagination
+    end
+
+    def allow_pagination=(allow_pagination)
+      @@allow_pagination = allow_pagination
     end
 
     def sandbox_mode=(sandbox_mode)
@@ -216,6 +225,14 @@ module Quickbooks
     def initialize(msg)
       self.message = msg
       super(msg)
+    end
+  end
+
+  class Configuration
+    def self.set_tokens_and_realm_id token, secret, realm_id
+      Thread.current[:token]    = token
+      Thread.current[:secret]   = secret
+      Thread.current[:realm_id] = realm_id
     end
   end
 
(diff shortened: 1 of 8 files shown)


lib/quickbooks-ruby.rb
```diff
@@ -219,4 +228,12 @@ def initialize(msg)
     end
   end
 
+  class Configuration
+    def self.set_tokens_and_realm_id token, secret, realm_id
+      Thread.current[:token]    = token
```
minimul: Why Thread.current?

lib/quickbooks/model/base_model.rb
```diff
@@ -12,6 +12,48 @@ def initialize(attributes={})
...
+      def self.all query = nil, options = {}
+        model_name           = self.name.split('::').last
+        query                = "SELECT * FROM #{model_name}" if query.blank?
+        service              = eval("Quickbooks::Service::#{model_name}").new
+        service.access_token = OAuth::AccessToken.new(
+          $qb_oauth_consumer, options[:token] || Thread.current[:token],
+          options[:secret] || Thread.current[:secret]
+        )
+        service.company_id   = options[:realm_id] || Thread.current[:realm_id]
+        service.query(query).entries
```
minimul: Can service.query() return nil?

lib/quickbooks/model/base_model.rb
```diff
@@ -12,6 +12,48 @@ def initialize(attributes={})
...
+      end
+
+      def self.where options = {}
+        model_name = self.name.split('::').last
```
minimul: Could use self::REST_RESOURCE since API doesn't care about case for entities but this is fine. Perhaps should be in its own method like `model` for reuse.

lib/quickbooks/model/base_model.rb
```diff
@@ -12,6 +12,48 @@ def initialize(attributes={})
...
+        query      = options.present? ? "SELECT * FROM #{model_name} WHERE " : "SELECT * FROM #{model_name}"
+        if options.is_a?(String)
+          conditions = options
+        else
+          conditions = options.collect do |key, value|
+            if value.is_a?(Array)
+              "#{key} IN ('#{value.join("','")}')"
```
minimul: `value` should leverage Quickbooks::Util::QueryBuilder e.g.
https://github.com/minimul/quickbooks-ruby-base/blob/master/lib/quickbooks/base/finders.rb

lib/quickbooks/model/base_model.rb
```diff
@@ -12,6 +12,48 @@ def initialize(attributes={})
         attributes.each {|key, value| public_send("#{key}=", value) }
       end
 
+      def self.all query = nil, options = {}
```
minimul: Use parenthesis for style consistency.

minimul: What is the difference between `.query_in_batches()`?

lib/quickbooks-ruby.rb
ruckus: My guess is that `Thread.current` is a safe place to store what is essentially treated as a global variable, without using standard Ruby global variables like `$secret` which pollute the whole namespace.

kashif-umair: In addition to @ruckus comment: Thread.current is useful when you expect that multiple threads can have different tokens. So we cannot use a global variable in that case.

lib/quickbooks/model/base_model.rb
kashif-umair: I don't think so. Do you have any specific case in your mind?

lib/quickbooks/model/base_model.rb
kashif-umair: `query_in_batches()` will continue working fine unless user set `allow_pagination` to false. If `allow_pagination` is set to false then no pagination will happen.

minimul:
Let me say that I like this PR. I have seen the same needs but decided to add similar methods to [quickbooks-ruby-base](https://github.com/minimul/quickbooks-ruby-base).

One thing that gem does better is easy switching between entities. For example, to check IDs from
employe and/or customer it is:
```
base = Quickbooks::Base.new(account, :customer)
>> base.find_by_id(55) # searches customer
>> base.find_by_id(55, :employee) # searches employee
```

I don't like to have to write out the entire namespace.

That said, I am personally very open to getting these methods in. I would like to dig in more to aid this cause, however, it might take a bit of time.

lib/quickbooks/model/base_model.rb
minimul: I'd have to test live but I thought that in some cases it returns nil and not just an empty array.

kashif-umair:
hi @minimul and @ruckus, I've been busy at office for this long time period that's why I couldn't follow up on this PR. Can we have this PR merged in or do you need to review it again :)

ruckus:
Hi @kashif-umair - thank you for this contribution. This looks pretty sweet.

@minimul - did you have any comments? 

Otherwise I dont see why we cannot get this merged in. Thanks again.

minimul:
Good by me.

ruckus:
Thank you!!

lib/quickbooks/model/active_record_scaffold.rb
```diff
@@ -0,0 +1,85 @@
...
+        def query_in_batches(options = {})
+          initialize_service(options).query_in_batches(query, options).entries
+        end
+
+        private
+
+        def initialize_service(options = {})
+          service              = eval("Quickbooks::Service::#{model}").new
+          service.access_token = OAuth::AccessToken.new(
+            $qb_oauth_consumer, options[:token] || Thread.current[:token],
```
ruckus: @kashif-umair where does `$qb_oauth_consumer` come from? This indicates its a global variable which needs to be set by the user. We'll need a better way to load this instance of `OAuth::Consumer` into the library.

I'm going to un-do this PR until this can be resolved as its currently in a non-functional state.

kashif-umair: Oh man, I was testing these changes in the product where we use this gem and we had a global variable `$qb_oauth_consumer` in that product code. That's why this implementation was working fine for me. Sorry about that. I'll send a fix within 24 hours.

minimul:
@kashif-umair Can you look at issue #301 as well.

Issue #270, opened 2015-10-28
Can't create non inventory Item - need minorversion=4


vanboom:
I am having trouble creating a non inventory item...

``` ruby
item = Quickbooks::Model::Item.new(:type=>Quickbooks::Model::Item::INVENTORY_TYPE, :name=>"TESTITEM", :description=>"TEST ITEM", :income_account_id=>income_account.id, :asset_account=>asset_account.id)
newitem.id = service.create(item)
```

The item is successfully created, but shows up in Quickbooks as a Service item.  After manually changing the item type in Quickbooks, I can query the item back into Ruby...

```ruby
fetched_item = service.query("select * from Item where Id = '%s'"%newitem.id)
puts fetched_item.first.type
```

The Type returned by the fetched item is "Service"  

I enabled logging, and it appears that the type field in the XML returned by Quickbooks is "Service".

I tried changing the type of the item in Quickbooks to "Inventory" and then the fetched item type is correctly set to "Inventory".  

Thanks for an awesome gem - I appreciate any help or guidance in how I can debug this further.

vanboom:
I created a Non Inventory item manually in Quickbooks, and then fetched the item by the ID number...

```xml
<IntuitResponse xmlns="http://schema.intuit.com/finance/v3" time="2015-10-28T11:53:54.598-07:00">
  <QueryResponse startPosition="1" maxResults="1">
    <Item domain="QBO" sparse="false">
      <Id>33</Id>
      <SyncToken>0</SyncToken>
      <MetaData>
        <CreateTime>2015-10-28T11:53:48-07:00</CreateTime>
        <LastUpdatedTime>2015-10-28T11:53:48-07:00</LastUpdatedTime>
      </MetaData>
      <Name>NONINVENTORYTEST</Name>
      <Description>non inventory test part</Description>
      <Active>true</Active>
      <FullyQualifiedName>NONINVENTORYTEST</FullyQualifiedName>
      <Taxable>true</Taxable>
      <UnitPrice>0</UnitPrice>
      <Type>Service</Type>
      <IncomeAccountRef name="parts">105</IncomeAccountRef>
      <PurchaseCost>0</PurchaseCost>
      <TrackQtyOnHand>false</TrackQtyOnHand>
    </Item>
  </QueryResponse>
</IntuitResponse>
```

This looks like a bug to me!

vanboom:
UPDATE: I found the answer on the Quickbooks API Help
[...]

PR #271, opened 2015-10-28
Add minorversion 4 support to Item service and model, reference #270

vanboom:
Without minorversion 4, Non-Inventory items are not supported for the Item object.

Reference issue #270.  Non inventory items will be returned from Quickbooks Online with the Type attribute set to "Service" even though in Quickbooks the item will appear as a "Non Inventory" item.

This PR implements setting minorversion=4 only for the Item service and model to enable support for Non Inventory items.

### DIFF
--- a/lib/quickbooks/model/item.rb
+++ b/lib/quickbooks/model/item.rb
@@ -12,7 +12,7 @@ module Quickbooks
       REST_RESOURCE = 'item'
 
       INVENTORY_TYPE = 'Inventory'
-      NON_INVENTORY_TYPE = 'Non Inventory'
+      NON_INVENTORY_TYPE = 'NonInventory'
       SERVICE_TYPE = 'Service'
       ITEM_TYPES = [INVENTORY_TYPE, NON_INVENTORY_TYPE, SERVICE_TYPE]
 
--- a/lib/quickbooks/service/item.rb
+++ b/lib/quickbooks/service/item.rb
@@ -7,6 +7,16 @@ module Quickbooks
         update(item, :sparse => true)
       end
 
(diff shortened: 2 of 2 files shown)


ruckus:
Thanks for the code. Couple questions:

* Should we allow an `options` hash to both those `url_for_` methods such that a user can specify a different `minorversion` if need be? In lieue of it being hard-coded at 4

* Would the version bump be to `0.4.1`? I'm not super up on semantic-versioning, so let me know if thats not correct 

Thanks again.

vanboom:
Yes, good call... I will add the options hash.  I am not sure what to make the version... is it common for the PR to update the version?  Or leave it alone? 

thanks!

ruckus:
> is it common for the PR to update the version? Or leave it alone?

Yes, I brain-farted on that one.

No, don't change it yourself. Actually, remove it from the PR if you can. I will manipulate that when I merge the PR and release a new version.

vanboom:
After some thought...
I am comfortable hard coding the minorversion parameter because it relates to the XSD/XML schema that is tied closely to the implementation of ../models/item.rb.   If the implementation of Item.rb changes in the future to support a later minorversion, then the minorversion would be bumped in the same way as the API endpoint v3. 

I will remove version from the PR.  Thanks!

ruckus:
Ok, then how about making it a constant in the `Item` model? Right now, that 4 is kind of buried in the code.

vanboom:
Done!  I also removed the version from the PR.

ruckus:
Thanks, released `0.4.1`

vanboom:
Awesome - thanks!

PR #275, opened 2015-11-11
correct fetch_by_id method url method

jordangraft:
Skip using the ```url_for_resource``` method to build the url for ```fetch_by_id``` so that you can include the ```minorversion``` param in other methods.


ruckus:
Thanks for this. However, any other service which uses `fetch_by_id` is still affected (?)

ruckus:
Der, my bad. The `minorversion` stuff was only for `Item`s anyway.

PR #276, opened 2015-11-11
added void method for service/payment

jordangraft:
[QBO Void Payment Docs](https://developer.intuit.com/docs/api/accounting/Payment)

One more.   Added a ```void``` method to the payment service so that you can utilize that QBO API functionality.


ruckus:
Sweet, thanks!

PR #290, opened 2016-01-28
add item type 'category' to valid item types

rolfat:
As of Nov 14, 2015 there is a new item type called 'Category'. See Intuit documentation here:
https://developer.intuit.com/docs/api/accounting/item

> In addition to the above, QuickBooks companies will start supporting item categories to define item hierarchies... ... Use Item.Type set to Category to create hierarchies. See Migrating to categories for integration information and see Category object definition and CRUD examples on this page.


ruckus:
Thank you!

rolfat:
Glad to contribute. Keep up the stellar work!

rept:
Is there a way to get the possible Category types?

PR #293, opened 2016-02-10
Remove dependency on alias_method_chain from create_http_request method.

dafstone:
I took a look at this with another member of my team and we felt that removing the dependency on the alias_method_chain was the best way to go here given the way the code was written.  This assumes compatibility with the existing create_http_request method, as we believe was the original intent of the code.  Passes tests.

### DIFF
--- a/lib/quickbooks/util/multipart.rb
+++ b/lib/quickbooks/util/multipart.rb
@@ -4,7 +4,7 @@ end
 require "oauth"
 
 OAuth::Consumer.class_eval do
-  def create_http_request_with_multipart(http_method, path, *arguments)
+  def create_http_request(http_method, path, *arguments)
     http_method = http_method.to_sym
 
     if [:post, :put].include?(http_method)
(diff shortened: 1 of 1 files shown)


ruckus:
Hi @dafstone thanks for the commits. I'm trying to remember the reason why I even implemented `alias_method_chain` in the first place, but sure enough if I use your commits and try an upload it works, as do other basic REST endpoints. It just scares me the fix was as simple as that! :)

Anyways - is there a reason why you closed this PR?

Thanks again.

dafstone:
Hi! I actually closed and reopened the PR to kick Travis into a rebuild to see if it would pass. I can't actually recreate the failures that Travis is reporting on my end with ruby 1.9.3 and ruby 2.2.0, and it looks like it might be something with the Travis build, and not the actual code itself - this PR should be reading as open.

dafstone:
Hi @ruckus was wondering if you were going to merge this in, your last note indicated that you might.  Thanks in advance!

PR #294, opened 2016-02-11
Add transfer support

Craggar:
I need transfer support for my product, so I've added model/service for transfers, and tests


ruckus:
Thank you!

I made some adjustments to the transfer model - namely changing the `to_account` attribute to `to_account_ref` (ditto for `from_account`) to match the conventions from other models. Also I added the `reference_setters` convenience method.

Thanks again.

Craggar:
Great, thanks. I think I'm a few commits behind (as we have to test things a bit before releasing, so I tend to be a little behind the master branch). I'll take a look at your changes, as I'm putting together the CompanyCurrency and ChangeDataCapture at the moment also.

ruckus:
Thanks! I just pushed `0.4.3`

PR #297, opened 2016-02-26
Use constant for qb_consumer instead of global variable.

lmatiolis:
Just a small change on README to suggest the usage of a constant instead of a global.

Globals should be avoided, as seen on ruby-style-guide.
Rubocop error: 'Do not introduce global variables.'


ruckus:
Thank you!

lmatiolis:
Thanks for the gem! :)

Issue #300, opened 2016-03-02
As_json and to_json


bendangelo:
```
      customers = service.all.first.as_json
```

Will go into an infinite loop and crash.

# /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:161:in `map'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:161:in `as_json'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:52:in `as_json'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:142:in `block in as_json'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:142:in `map'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:142:in `as_json'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:161:in `block in as_json'
     # /Users/ben/.rvm/gems/ruby-2.3.0@braavos/gems/activesupport-5.0.0.beta2/lib/active_support/core_ext/object/json.rb:161:in `each'

minimul:
Any ideas of how to fix?

bendangelo:
I just did this. 
`json = Hash.from_xml(found_invoice.to_xml.to_s).to_json`

minimul:
Closed in https://github.com/ruckus/quickbooks-ruby/pull/302

shir:
Looks like this issue appear again. I've updated to latest version 0.6.1 from 0.4.2 and get this issue with infinity `as_json`

PR #302, opened 2016-03-12
as_json and to_json throwing stack too deep exception   #300

muhammad-abubakar:
`to_json` or `as_json` on model  gives stack too deep exception. Reason behind this is BaseModel contains a ROXML instance variable `roxml_references` which itelf has a model object and thus it goes into infinite recursion. Solutions is to exclude `roxml_references` while calling `as_json` because it only contains gem/system level information.


muhammad-abubakar:
@minimul  @ruckus  can you please take a look at this?

minimul:
Thanks!

PR #305, opened 2016-03-29
Fix duplicate spec name.

lmatiolis:
It was saying it shouldn't create with an empty employee_ref but was testing for vendor_ref.


minimul:
Appreciated!

PR #306, opened 2016-04-05
don't go :boom: when a NameValue pair is missing

UnderpantsGnome:
I stumbled on this when using sandbox data. When calling company_info.industry_code for some reason that pair wasn't in the data and it throws a nil exception.

Wasn't sure if that's the desired behavior, so here's a fix if it's not.


ruckus:
Awesome, thank you!!

UnderpantsGnome:
You're welcome. Thanks for doing all the tedious work :+1:

PR #307, opened 2016-04-07
Support Void for Invoice & SalesReceipt

jozefvaclavik:
This adds VOID method for Invoice & SalesReceipt. Pretty much same as Void for Payment model.


minimul:
Thanks Jozef.

PR #311, opened 2016-04-13
Entity on a DepositLineDetail should be of type BaseReference

kurabi-hc:
As per the Intuit developer docs. An entity is basically a BaseReference.

Link to [docs](https://developer.intuit.com/docs/api/accounting/Deposit) and [screenshot](https://www.dropbox.com/s/mbp2raoujzinmzw/Screenshot%202016-04-11%2013.48.03.png?dl=0) attached. I also verified this and it works.


minimul:
Thanks. Yep. Thanks for catching this.

PR #316, opened 2016-04-22
Add support for Change Data Capture

Craggar:
Fetches the Change Data Capture since a given date. Also included an all_types function on the model that parses the XML into Quickbooks::Model::{entity} objects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: ruby
 rvm:
   - 1.9.3
-  - 2.0.0
-  - 2.1.0
-  - 2.2.0
+  - 2.2.3
 
 sudo: false
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: ruby
 rvm:
   - 1.9.3
-  - 2.2.3
+  - 2.0.0
+  - 2.1.0
+  - 2.2.0
 
 sudo: false
 notifications:
--- a/README.md
+++ b/README.md
@@ -561,8 +561,30 @@ end
 
 ## Change Data Capture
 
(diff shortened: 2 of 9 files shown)


minimul:
I agree with this PR as the current approach https://github.com/ruckus/quickbooks-ruby#change-data-capture doesn't support query multiple entities in one request.

Thanks.

@Craggar Can you also add some examples to the README.

Craggar:
@minimul I'll get onto some examples for the readme, for sure. Probably won't be until the end of the week or the weekend, though, but I'll get round to it!

Craggar:
I've added some examples to the readme, and also made a change to parse deleted entities as 'ChangeModels' as previously they were just parsed as their type (e.g. Quickbooks::Model::Invoice), which lost the 'Deleted' status flag.

minimul:
Excellent – failing on Travis however. https://travis-ci.org/ruckus/quickbooks-ruby/builds/128333976

Craggar:
It appears to be failing during the bundle install of the listen gem - looks like it affected the other pull request too (update spec helper). I think that's a problem in Travis, though, as I haven't modified the gemspec file.

minimul:
Would you remove 2.0.0 and 2.1.0 and replace 2.2.0 with 2.2.3 in travis.yml and then push?

Craggar:
The Ruby 1.9.3 build still fails - looks like Listen was updated recently, and doesn't support anything prior to 2.2.x.

minimul:
In 1.9.3's case it is ruby_dep but I am sure the listen gem is also going to fail. These deps are from guard. @ruckus We should remove guard and guard-rspec because they are individualistic choices for a dev and not required to run the specs. Personally, I don't use guard to run specs and the choice should be up to each dev.

ruckus:
I dont even remember what guard is and/or why it was added at all. Lets remove it!

minimul:
@Craggar If you could also remove `guard` and `guard-rspec` from `quickbooks-ruby.gemspec`.

Craggar:
👍

PR #320, opened 2016-05-31
Allow an invoice to be voided using only the Id and SyncToken

chrismhilton:
I've made this change to the `Service::Invoice#void` method, that was added by @jozefvaclavik via 339d11765318d14f8a6e41e8dc9aa80965f9e34f, so that it's possible to void an invoice by only passing its `Id` and `SyncToken` values.

The method now uses a URL containing `operation=void` as per the [documentation](https://developer.intuit.com/docs/api/accounting/Invoice) so it's more like the `Service::ServiceCrud#delete_by_query_string` method. Removing the call to `invoice.valid?` means that you no longer have to use a `Model::Invoice` object with a customer reference and at least one line item present.

### DIFF
--- a/lib/quickbooks/service/invoice.rb
+++ b/lib/quickbooks/service/invoice.rb
@@ -24,11 +24,10 @@ module Quickbooks
       end
 
       def void(invoice, options = {})
-        raise Quickbooks::InvalidModelException.new(invoice.errors.full_messages.join(',')) unless invoice.valid?
-        xml = invoice.to_xml_ns(options)
-        url = "#{url_for_resource(model::REST_RESOURCE)}?include=void"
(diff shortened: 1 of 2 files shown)


ruckus:
Great, thank you!!

chrismhilton:
Many thanks for the merge. Is there an ETA for a new release?

ruckus:
@chrismhilton I'm preparing a release now and am putting in usage notes for this new void operation in the HISTORY. Do you know if `SyncToken` is required for the void operation - or is only `id` sufficient?

chrismhilton:
Thanks for the update @ruckus. You have to pass both the `Id` and `SyncToken` and everything else is optional.

ruckus:
Released!

https://github.com/ruckus/quickbooks-ruby/blob/master/HISTORY.md#044-2016-06-02

PR #330, opened 2016-09-17
Introduce ActiveRecord like methods to Quickbooks Models

kashif-umair:
### This PR is an improvement to previous PR that was merged but then reverted because of an issue. The previous PR is https://github.com/ruckus/quickbooks-ruby/pull/264

- This PR introduces some methods i.e. `all, where, find_by, find, count` methods to quickbooks models. This change eliminates the requirement of using services to query something from Quickbooks. You can directly call `Quickbooks::Models::Invoice.all` to get all the invoices.
- ~~Another setting `allow_pagination` is introduced in this PR. This is a boolean setting which is set to `true` by default. If user sets this setting to `false` then no pagination will be applied and all the records will be returned on any query.~~

---
- The golbal setting `allow_pagination` has been removed which I introduced in my last commit. Instead of that, I've introduced another option to skip pagination on each request basis i.e. `skip_pagination`. This setting will be available in both the conventional query method and the new methods introduced in this PR.

- [x] I'm working on updating the `README.md` file.


README.md
```diff
@@ -175,7 +181,41 @@ end
...
+Quickbooks::Model::Customer.all(nil, skip_pagination: true) #=> Returns array of all customers
+```
+The first argument in the method is query and if you don't pass any query then the default query will be executed which is, in this case, `SELECT * FROM Customer`
+
+Quickbooks models behave exactly the same like ActiveRecord for querying purposes. That means that these models have other methods e.g: `where` `count` `find_by` `find` etc. Following are some examples:
+```ruby
+
+# where
+Quickbooks::Model::Customer.where('') #=> runs default query and returns first 20 customers
+Quickbooks::Model::Customer.where(id: 170) #=> runs a conditional query and returns first 20 customers having id equal to 170
```
drewish: This doesn't seem like a great example because you shouldn't have multiple customers with the same id.

README.md
```diff
@@ -175,7 +181,41 @@ end
...
+Quickbooks::Model::Customer.where(id: 170, skip_pagination: true) #=> runs the conditional query and returns all customers having id equal to 170 without pagination
```
drewish: Same here, would be better to use a more realistic value to demonstrate the filtering. I'm not clear what "without pagination" means? The pagination will happen behind inside that method call right?

README.md
```diff
@@ -175,7 +181,41 @@ end
 
 ## Getting Started - Retrieving a list of Customers
 
-The general approach is you first instantiate a `Service` object based on the entity you would like to retrieve. Lets retrieve a list of Customers:
+There are two approaches to retrieve results from Quickbooks. However, you're encouraged to use the first approach as we'll be removing the second approach in the long run.
+
+###### Approach 1
```
drewish: This seems like it should be a heading level 3, e.g. `###`

lib/quickbooks/model/active_record_scaffold.rb
```diff
@@ -0,0 +1,113 @@
...
+            all(build_string_query(options), {})
+          elsif options.present?
+            all(build_hash_query(options.except(:skip_pagination)), options)
+          else
+            all(full_query, {})
+          end
+        end
+
+        def find_by(options = {})
+          where(options).first
```
drewish: seems like this should do a limit 1 to avoid fetching and parsing records that aren't used.

lib/quickbooks/model/active_record_scaffold.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+      end
+
+      module ClassMethods
+        def all(query = nil, options = {})
+          initialize_service(options).query(query, options).entries
+        end
+
+        def first
+          all.first
```
drewish: should this really be calling all? seems wasteful if you just want one record

lib/quickbooks/model/active_record_scaffold.rb
```diff
@@ -0,0 +1,113 @@
...
+      def save(options = {})
+        if id.blank?
+          self.class.send(:initialize_service).create(self, options)
+        else
+          self.class.send(:initialize_service).update(self, options)
+        end
+      end
+
+      module ClassMethods
+        def all(query = nil, options = {})
```
drewish: i'm not sure this quite matches up with active record: http://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-all

PR #336, opened 2016-10-24
Add support for line item groups, aka: Bundle

florinpatrascu:


ruckus:
This looks really good! Thank you so much!

florinpatrascu:
Thank you, for accepting this PR. One of these days I'll try to work on the Estimate model too, for continuing to adding the support for the Bundles, unless somebody else will be quicker :)

Issue #92, opened 2014-04-19
Hardy Service


markrickert:
Hey there, so I'm working with this gem heavily now and I'm wondering if it would make sense to implement some kind of "hardy service" like the [`Harvested`](https://github.com/zmoazeni/harvested/blob/master/lib/harvest/hardy_client.rb#L46-L77) gem has... essentially instead of throwing an error when the server rate-limits the calls, it would detect the rate limiting and wait the appropriate amount of time before calling the same request over again.

Thoughts on a mechanism like this?

minimul:
I can't see how we would implement rate-limiting sanely in the service code at this point. 

@markrickert Thanks for detailing the problem and your solution.

PR #339, opened 2016-10-28
Raise ThrottleExceeded when rate limited

### LINKED ISSUE #338: Include more information in Quickbooks::Forbidden exception
I'm running into the rate limiting with the accounting API and it too me way to long to figure out that  `Quickbooks::Forbidden` meant I was being throttled. 

It seems like we could provide a more detailed explanation in the exception based on what's in the response body:

``` xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<IntuitResponse time="2016-10-27T15:07:56.120-07:00" xmlns="http://schema.intuit.com/finance/v3">
    <Fault type="SERVICE">
        <Error code="3001">
            <Message>message=ThrottleExceeded; errorCode=003001; statusCode=403</Message>
            <Detail>The request limit was reached.  Try again later</Detail>
        </Error>
    </Fault>
</IntuitResponse>
```

The existing `parse_intuit_error` method parses it correctly:

``` rb
{:message=>"message=ThrottleExceeded; errorCode=003001; statusCode=403",
 :detail=>"The request limit was reached.  Try again later",
 :type=>"SERVICE",
 :code=>"3001",
 :element=>"3001"}
```

So it would just seem to be a matter of passing it back.

drewish:
Fixes #338. 

Let me know what you think about this. It should be totally backwards compatible but would give a easy way to retry when rate limited. Here's how I'm using it:

``` rb
  def save_with_throttling qbo_record
    tries ||= 3
    nap_time = rand(15..30)

    connection.quickbooks.save!(qbo_record)
  rescue Quickbooks::ThrottleExceeded
    if (tries -= 1) > 0
      logger.info "Hit the rate limit, napping for #{nap_time} seconds"
      sleep nap_time
      retry
    else
      logger.info "Made 3 tries to avoid the rate limit. Aborting"
      raise
    end
  end
```

It's similar to what was suggested in #92 but won't retry on unrelated errors.

### DIFF
--- a/quickbooks-ruby.gemspec
+++ b/quickbooks-ruby.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'oauth', '0.4.7'
   gem.add_dependency 'roxml', '3.3.1'
   gem.add_dependency 'nokogiri'  # promiscuous mode
-  gem.add_dependency 'activemodel', '< 5.0.0'
+  gem.add_dependency 'activemodel', '< 6.0.0'
   gem.add_dependency 'multipart-post' # promiscuous mode
 
   gem.add_development_dependency 'rake', '10.1.0'
--- a/quickbooks-ruby.gemspec
+++ b/quickbooks-ruby.gemspec
@@ -17,7 +17,12 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'oauth', '0.4.7'
   gem.add_dependency 'roxml', '3.3.1'
   gem.add_dependency 'nokogiri'  # promiscuous mode
-  gem.add_dependency 'activemodel', '< 6.0.0'
+  # ActiveModel 5.0 requires ruby 2.2.2 or greater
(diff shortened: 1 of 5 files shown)


quickbooks-ruby.gemspec
```diff
@@ -17,7 +17,13 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'oauth', '0.4.7'
   gem.add_dependency 'roxml', '3.3.1'
   gem.add_dependency 'nokogiri'  # promiscuous mode
-  gem.add_dependency 'activemodel', '< 5.0.0'
+  # ActiveModel 5.0+ requires ruby 2.2.2+
+  version = RUBY_VERSION.split('.').map(&:to_i)
+  if version[0] >= 2 && version[1] >= 2 && version[2] >= 2
```
drewish: I'm open to less tacky ways to do this but it got me unblocked.

minimul:
Thanks @drewish We will get to this soon.

ruckus:
Thank you!

Issue #256, opened 2015-07-21
Request ID QuickBooks


dheeraj510:
Hello,

I found some interesting blog post here https://developer.intuit.com/blog/2015/04/06/15346
It says by using Request ID data is not duplicated in the event of a disruption. this is what exactly I am looking for. I could not figure out how to send Request ID while posting journal entry or sales receipt to stop duplicating same data if posted again and again.

can we use this from  quickbooks-ruby Gem? if possible can you provide some sample code for journal entry or sales receipt?

Thank you.

ruckus:
Hey @dheeraj510  - you want to give this a whack? Used like:

```ruby
created_invoice = @service.create(invoice, :query => {:requestid => "foobar123"})
```

dheeraj510:
Hi @ruckus, Thanks for responding quickly.  I attempted to do it with journal entry using above syntax.  But it seems request id is not sent in my request and journal entries are duplicated in my quickbook account. Is request_id can be used only in invoice or it can be used also for journal entry and sales receipt too?

Following is my code for journal entire:

service = Quickbooks::Service::JournalEntry.new({access_token: oauth_client, company_id: company_id})
journal_entry = Quickbooks::Model::JournalEntry.new(:doc_number => "test", :txn_date => Date.today, :private_note => "test" )
journal_entry.line_items = []

line = Quickbooks::Model::Line.new
line.line_num = 1
line.detail_type = 'JournalEntryLineDetail'
line.description = "test"
line.amount = 12
jeld = Quickbooks::Model::JournalEntryLineDetail.new
jeld.posting_type = 'Credit'
jeld.account_id = 334
line.journal_entry_line_detail = jeld
journal_entry.line_items << line

line = Quickbooks::Model::Line.new
line.line_num = 2
line.detail_type = 'JournalEntryLineDetail'
line.description = "test"
line.amount = 12
jeld = Quickbooks::Model::JournalEntryLineDetail.new
jeld.posting_type = 'Debit'
jeld.account_id = 345
line.journal_entry_line_detail = jeld
journal_entry.line_items << line

service.create(journal_entry , :query => {:requestid => "8f199EGdg5dASY1fG5bEzkt5k2cNgBxx"})

 ------ QUICKBOOKS-RUBY REQUEST ------

METHOD = post
RESOURCE = https://sandbox-quickbooks.api.intuit.com/v3/company/1315124650/journalentry
REQUEST BODY:
[...]

PR #344, opened 2016-11-21
Pass query params (e.g. requestid) with batch

### LINKED ISSUE #343: Dealing with Net::ReadTimeout during batch save
I've been running into some `Net::ReadTimeout` exceptions when saving batches of records. I'm not sure what the default timeout is but with 30 records it happens pretty frequently, dropping it down to 10 records cuts it down considerably. 

It looks like a simple retry strategy will result in duplicate records (I was surprised to learn that QBO doesn't seem to have many unique indexes). I'm curious what folks do to deal with this. Do you catch exceptions then search for the submitted values to see if any records saved? Anyone spent time figuring out how to extend the timeouts in the OAuth client?

drewish:
Fixes #343 

I copied the gist of the spec from #256 but let me know what you think.

### DIFF
--- a/lib/quickbooks/service/batch.rb
+++ b/lib/quickbooks/service/batch.rb
@@ -3,7 +3,7 @@ module Quickbooks
     class Batch < BaseService
 
       def make_request(entity, options = {})
-        response = do_http_post(url_for_resource('batch'), valid_xml_document(entity.to_xml_ns))
+        response = do_http_post(url_for_resource('batch'), valid_xml_document(entity.to_xml_ns), options[:query])
         Quickbooks::Model::BatchResponse.from_xml(response.plain_body)
       end
     end
--- a/spec/lib/quickbooks/service/batch_spec.rb
+++ b/spec/lib/quickbooks/service/batch_spec.rb
@@ -17,4 +17,12 @@ describe "Quickbooks::Service::Batch" do
       @service.make_request(Quickbooks::Model::BatchRequest.new)
     }.should raise_error(Quickbooks::IntuitRequestException)
   end
+
+  it "allows user to specify a RequestId in a create call" do
+    requestid = "foobar123"
+    xml = fixture("batch_response.xml")
+    stub_request(:post, "#{@service.url_for_resource('batch')}?requestid=#{requestid}", ["200", "OK"], xml)
+    batch_resp = @service.make_request(Quickbooks::Model::BatchRequest.new, :query => {:requestid => requestid})
+    batch_resp.class.should == Quickbooks::Model::BatchResponse
+  end
 end


drewish:
Humm... testing this out manually I'm not sure it's working. The second time I try making the request I get a super generic error with `response.response_items.first.fault.errors` is showing:
```
  @code="6000",
  @detail="Business Validation Error: An unexpected error occurred while accessing or saving your data. Please wait a few minutes and try again.  If the problem persists, contact customer support.",
  @element="",
  @message="A business validation error has occurred while processing your request",
```
Their [blog post about requestids](https://developer.intuit.com/hub/blog/2015/04/06/15346) make it sound like it should just work:
> For batch operations, the RequestID supports a maximum of 36 characters. For each batch ID, only 10 characters are allowed when RequestID is also specified.

I'll reach out to their support.

minimul:
I am using `SecureRandom.uuid` for the requestid in [qbo_api](https://github.com/minimul/qbo_api/blob/master/lib/qbo_api/util.rb) not that there should be anything wrong with `SecureRandom.urlsafe_base64` but who knows.

Let me see the log for the batch request.

drewish:
It gets stranger. I decided to keep testing this with other requests in my sync process and it seems like that only happens with batches of Items. I'm able to re-request customers and invoices with no issues.

drewish:
I'll try SecureRandom.uuid and see if that makes any difference. I don't remember how I got to urlsafe_base64.

drewish:
Here's logs for a batch of items and a batch of customers: [logs.zip](https://github.com/ruckus/quickbooks-ruby/files/604773/logs.zip)

The customers behave oddly in a different way. For those when I make the second request I get back "Stale Object Error : You and andrew morton were working on this at the same time. andrew morton finished before you did, so your work was not saved."

minimul:
The only thing that I see is that your sparse attribute shouldn't be at the root level e.g. `<IntuitBatchRequest xmlns="http://schema.intuit.com/finance/v3" sparse="false">` but instead at the batch entity level. From API docs

```xml
<IntuitBatchRequest xmlns="http://schema.intuit.com/finance/v3">
     <BatchItemRequest bId="bid1" operation="create">
         <Vendor>
             <DisplayName>Smith Family Grocery</DisplayName>
         </Vendor>
     </BatchItemRequest>
     <BatchItemRequest bId="bid2" operation="delete">
         <Invoice>
             <Id>119</Id>
             <SyncToken>0</SyncToken>
         </Invoice>
     </BatchItemRequest>
     <BatchItemRequest bId="bid3" operation="update">
         <SalesReceipt xmlns="http://schema.intuit.com/finance/v3" sparse="true">
             <Id>38</Id>
             <SyncToken>0</SyncToken>
             <PrivateNote>This is another private note.</PrivateNote>
         </SalesReceipt>
     </BatchItemRequest>
         <BatchItemRequest bId="bid4">
         <Query>select * from SalesReceipt where TotalAmt > '200.00'</Query>
    </BatchItemRequest>
 </IntuitBatchRequest>
```

drewish:
Response from the first round of Intuit support:
> This looks like a not supported feature to me for Name List entities.

> I believe you cannot use RequestId for any name list entities as of now.
I have raised a request to engineering team to analyze this more.
But from the logs that I see, it seems you cannot use requestid for name list the same way as you do for Txns.

> I would suggest you to instead use Webhooks or CDC(Changed Data Capture) to get the changed data-

drewish:
Since this works correctly for invoices I think it's worth merging. Hopefully Intuit will either fully implement the functionality or at least update the docs to make it clear which entity types are supported.

drewish:
@minimul any other feedback on this PR?

minimul:
fwwi: I commented on Intuit's response briefly in my recent newsletter http://minimul.com/qr-newsletter/issue-22.html

drewish:
I forgot to put the rest of my exchange with them up here. Here was my reply:

> That seems like a pretty big restriction on the functionality that is not documented. Please let me know what you hear from the engineering team. At the very least you should update the documentation to make it clear what types of requests support the requestid parameter.

And their response:

> Yes Andrew, I agree.
> We have asked the engineering team to confirm. Thsi migth take 3-4 weeks and then I will be able to update you accordingly and get the docs updated.

So I'll follow up if I hear more.

drewish:
Here's the final word from Intuit:
> As of date RequestId for batches works only for Txn entities.
> I will get the docs updated.

PR #348, opened 2016-12-08
Added exception handling for new 429 too many requests throttling.

stevedev:
QBO has added 429 errors for the new throttling they have implemented. Not sure if it is active in production yet, but from what we've seen it is in sandbox so we were able to generate the error properly. 

Here is a blog post with the announcement from August 8, 2016: 
https://developer.intuit.com/hub/blog/2016/08/08/idg-concurrent-api-use-changes-coming-soon

This ticket adds the 429 handling (with a spec).

### DIFF
--- a/spec/fixtures/too_many_requests_error.xml
+++ b/spec/fixtures/too_many_requests_error.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<IntuitResponse time="2016-12-07T15:27:09.527-08:00" xmlns="http://schema.intuit.com/finance/v3">
+  <Fault type="SERVICE">
+    <Error code="3001">
+      <Message>message=ThrottleExceeded; errorCode=003001; statusCode=429</Message>
+      <Detail>The request limit was reached.  Try again later</Detail>
+    </Error>
+  </Fault>
+</IntuitResponse>
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -214,20 +214,13 @@ module Quickbooks
     end
   end # << self
 
-  class InvalidModelException < StandardError;
-  end
-
-  class AuthorizationFailure < StandardError;
-  end
-  class Forbidden < StandardError;
-  end
-  class ThrottleExceeded < Forbidden;
-  end
-
-  class ServiceUnavailable < StandardError;
-  end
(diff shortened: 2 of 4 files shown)


minimul:
Perfect. Thanks.

drewish:
Thanks this was driving me crazy. Seems like a good reason to get a new release out.

spec/fixtures/too_many_requests_error.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<IntuitResponse time="2016-12-07T15:27:09.527-08:00" xmlns="http://schema.intuit.com/finance/v3">
+  <Fault type="SERVICE">
+    <Error code="3001">
+      <Message>message=ThrottleExceeded; errorCode=003001; statusCode=429</Message>
```
drewish: So odd to me that we introduced a new exception when they use the same message internally. What do you think about just having this rase a ThrottleExceeded instead of adding another thing to rescue?

stevedev: I actually thought it was odd that the message was *Throttle Exceeded* when in their blog post Intuit specifically said the message would be *Too Many Requests*. It really should be Too Many Requests if they are using a 429 response, at least according to this RFC: https://tools.ietf.org/html/rfc6585#section-4 and I'd argue that they are actually getting it wrong here. Of course, them getting it wrong doesn't mean we shouldn't just go with *Throttle Exceeded* since it's possible that they might just keep it wrong until the next version of the API is released.

Note that they are also deprecating the 503, which was a fairly unhelpful code to use for this in the first place since it actually should have been returning a *Service Unavailable* message and not *Throttle Exceeded*. I also think that you should probably be handling a 503 differently than a 429 - 503 should mean that the service is unavailable and you should retry using something like an exponential backoff, whereas for a 429 you could just retry a minute later, but that's just my opinion on that. On the other hand, they just seem to start redirecting with 403 responses when the API is actually unavailable so who knows. *sigh*

Happy to discuss this further. I'm actually also curious if anybody here has been receiving 429's yet in production. I haven't seen any yet... still seeing 503s.

drewish: Just been seeing them from sandbox sites. Your logic makes sense to me. Just wanted to make sure it was a deliberate choice.

PR #354, opened 2016-12-29
all should always return an array

### LINKED ISSUE #353: `all` returns `nil` rather than empty array for no results
I had some code that looked like:
```rb
records = service(type).all(query)
records.each do |record|
  ...
end
```
It was blowing up with an "NoMethodError: undefined method `each' for nil:NilClass" error.

A little poking shows that `.all` returns `nil` rather than the empty array I'd expect if the query returns no results:
```rb
irb(main):029:0> con.quickbooks.service(:customer).query("select * from customer where id = '-1'")
=> #<Quickbooks::Collection:0x007fb89a1a2638 @count=0, @entries=[]>
irb(main):030:0> con.quickbooks.service(:customer).query("select * from customer where id = '-1'").entries
=> []
irb(main):031:0> con.quickbooks.service(:customer).all("select * from customer where id = '-1'")
=> nil
```

The issue seems to be in the [implementation of `all`](https://github.com/ruckus/quickbooks-ruby/blob/master/lib/quickbooks/service/service_crud.rb#L15):
```rb
      def all(object_query=nil, options={})
        collection = []
        self.query_in_batches(object_query, options) do |batch|
          collection << batch.entries
        end
        collection.flatten!
      end
```

Looking at the ruby docs, I learned that in addition to modifying the array in place, `flatten!`:
> Returns nil if no modifications were made (i.e., the array contains no subarrays.)
http://ruby-doc.org/core-2.2.0/Array.html#method-i-flatten-21

So I'll open up a PR to change that to `flatten` and provide a less surprising result.

drewish:
`flatten!` returns `nil` if no modifications are made. So a query that returns no results would return `nil` rather than `[]`.

Fixes #353

### DIFF
--- a/lib/quickbooks/service/service_crud.rb
+++ b/lib/quickbooks/service/service_crud.rb
@@ -12,7 +12,7 @@ module Quickbooks
         self.query_in_batches(object_query, options) do |batch|
           collection << batch.entries
         end
-        collection.flatten!
+        collection.flatten
       end
 
       def query_in_batches(object_query=nil, options={})
--- a/spec/lib/quickbooks/service/service_crud_spec.rb
+++ b/spec/lib/quickbooks/service/service_crud_spec.rb
@@ -46,11 +46,21 @@ module Quickbooks
         ServicedClass.find_by(:name, "test")
       end
 
-      it "finds all" do
-        results = double(Quickbooks::Collection, count: 1)
-        results.should_receive(:entries).and_return []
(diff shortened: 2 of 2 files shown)


drewish:
Test failures look like some kind of a dependency issue with nokogiri and older versions of ruby.

ruckus:
Duplicate of #358 which was just merged.

PR #356, opened 2017-01-03
Enhance the token renewal example

### LINKED ISSUE #355: Enhance token renewal example
My first pass at the token renewal code was based off what's in the readme. It assumes an `error_code="0"` so if you hit the `"212"` you'll overwrite valid credentials with nil values. We should add some error handling hints.

There were a couple of other things I'd like to tweak:
- Date computation uses 5 and 6 months but Intuit says the token is valid for 180 days and can be renewed 30 days ahead. I think it'd be better to go of those values rather than the vaguer "months".
- It suggests storing two date fields but you really only need token_expires_at, you can just factor the 30 days into the query to find expiring credentials. 

I'll open up a PR for this next year. Mostly just putting this in as a reminder for myself.

drewish:
- Align the dates with Intuits documented values of 180 and 150 days
- Use a single date field to track expiration
- Examine error code when updating saved credentials

Fixes #355

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,22 +134,19 @@ access_token = OAuth::AccessToken.new(QB_OAUTH_CONSUMER, string_access_token_fro
 Most likely you will want to persist the OAuth access credentials so you don't have to connect to QBO
 each and every time.
 
-QBO allows the access credentials to live for 6 months, and after 5 months they can be renewed. You will
-get an error if you try and renew prior to 5 months. Thus, you will need to keep track of the dates and
-manage the renewal process yourself.
+The access credentials are valid for 180 days. 30 days before they expire, they can be renewed. You'll need
+to keep track of the expiration date and manage the renewal process yourself.
 
 An example database table would have fields likes:
 
-
 ```sql
 access_token varchar(255),
-access_secret varchar(255)
+access_secret varchar(255),
 company_id varchar(255),
-token_expires_at datetime # Set to 6.months.from_now upon insertion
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -134,38 +134,51 @@ access_token = OAuth::AccessToken.new(QB_OAUTH_CONSUMER, string_access_token_fro
...
+The access credentials are valid for 180 days. 30 days before they expire, they can be renewed. You'll need
+to keep track of the expiration date and manage the renewal process yourself.
 
 An example database table would have fields likes:
 
-
 ```sql
 access_token varchar(255),
-access_secret varchar(255)
+access_secret varchar(255),
```
drewish: I think it might be worth pointing out that storing the credentials in the clear like this violates Intuit's [guidelines for encrypting the access token, realmId, and access token secret](https://developer.intuit.com/docs/0100_quickbooks_online/0100_essentials/0085_develop_quickbooks_apps/0006_publish_and_market_your_app_with_quickbooks/0005_security_requirements). It probably only matters if you're trying going to be trying to publish an app in their marketplace.

ruckus: This is a valid point and you're absolutely right. Rather than confuse the example by implementing encryption perhaps we should just link to the page that you did above and say "its up to you to encrypt... so take what you're seeing in this README as purely a guideline" and then perhaps also link to something like https://github.com/attr-encrypted/attr_encrypted

drewish: Yeah agreed that it would be too confusing to include encryption in the example. I'll add a note about the need to encrypt though.

Issue #352, opened 2016-12-28
Can't bundle 0.4.6 with Rails 5


drewish:
Now that 0.4.6 has been released I tried to upgrade my Rails 5 app to use it. Previously I'd been pulling in the contents of master:
```rb
gem 'quickbooks-ruby', git: 'git@github.com:ruckus/quickbooks-ruby.git'
```
So I switched that to:
```rb
gem 'quickbooks-ruby', '~> 0.4.6'
```
But when I try to bundle I'm getting:
```
Bundler could not find compatible versions for gem "activemodel":
  In snapshot (Gemfile.lock):
    activemodel (= 5.0.0.1)

  In Gemfile:
    rails (~> 5.0.0) was resolved to 5.0.0.1, which depends on
      activerecord (= 5.0.0.1) was resolved to 5.0.0.1, which depends on
        activemodel (= 5.0.0.1)

    rails (~> 5.0.0) was resolved to 5.0.0.1, which depends on
      activerecord (= 5.0.0.1) was resolved to 5.0.0.1, which depends on
        activemodel (= 5.0.0.1)

    quickbooks-ruby (~> 0.4.6) was resolved to 0.4.6, which depends on
      activemodel (< 5.0.0)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

I'd hoped that the changes in https://github.com/ruckus/quickbooks-ruby/pull/339/commits/7ed9b701c93b7877f61152dfb54fc35951b2dd0c would solve this. It looks like the activemodel version is fixed according to the ruby version used to build the gem. 

I'm wondering if there's a way we can remove the explicit versioned dependency on ActiveModel?

drewish:
Here's a workaround:
```
# Use git until https://github.com/ruckus/quickbooks-ruby/issues/352 is resolved
gem 'quickbooks-ruby', '~> 0.4.6', git: 'git@github.com:ruckus/quickbooks-ruby.git'
```

drewish:
Hum I'm not sure why this didn't close

PR #357, opened 2017-01-03
Fix bundler versioning

drewish:
Trying out a fix for #352 based on this comment: https://github.com/bundler/bundler/pull/4651#issuecomment-268832706

I was able to fix my app's issues by adding `ruby RUBY_VERSION` to the Gemfile as suggested. Curious to see if that'll fix the issues with Travis too.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'http://rubygems.org'
 
+ruby RUBY_VERSION
+
 gem 'pry'
 
 gemspec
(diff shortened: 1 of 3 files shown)


drewish:
@ruckus / @minimul I hope you don't mind the ping but I'd love your feedback on this. It seems like a pretty clean way to get the versioning issues sorted out.

Gemfile
```diff
@@ -1,5 +1,7 @@
 source 'http://rubygems.org'
 
+ruby RUBY_VERSION
```
drewish: This should be ignored by previous versions of bundler.

ruckus:
Thanks! I didnt know about the `ruby` keyword

PR #358, opened 2017-01-04
Service.all returns nil if no elements exist, should return []

vanboom:
The flatten! method converts an empty array to nil.  Use flatten to return [] instead.


drewish:
Duplicate of https://github.com/ruckus/quickbooks-ruby/pull/354

ruckus:
Thank you!

PR #366, opened 2017-03-09
Fix reference setters for fields without _ref

cohendvir:
This is a fix for my previous commit

1. replace only the last occurrence of `_ref`, and non if not found
2. append `_id` as a name, even if `_ref` is not found

### DIFF
--- a/lib/quickbooks/model/base_model.rb
+++ b/lib/quickbooks/model/base_model.rb
@@ -86,7 +86,8 @@ module Quickbooks
           references = args.empty? ? reference_attrs : args
 
           references.each do |attribute|
-            method_name = "#{attribute.to_s.gsub('_ref', '_id')}=".to_sym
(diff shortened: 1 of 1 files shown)


ruckus:
I'm not 100% sure whats happening here, but this sounds like it might be an issue in your calling code? Updating the gem itself with logic to check for "last" kind of stuff has a bad smell.

Plus, the test suite is failing. For all PRs you'll want to run the test suite locally

cohendvir:
@ruckus, my previous pr didn't work because when setting a reference with `reference_setters`, its expecting the referenced field to end with `_ref`. This pr fixes that.
Also, in the current code, it replaces all occurrences of `_ref` to `_id`, meaning `field_refactor_ref` would change to `field_idactor_id`. I fixed to replace the last `_ref` only.

Tests are passing now

PR #377, opened 2017-05-22
close #375, improve consistency of email address setter method betwee…

### LINKED ISSUE #375: Suggestion:  Email discrepancy between Invoice and Sales Receipt
The billing email address handling is different between sales receipts and invoices:

Invoice uses "billing_email_address = "
Sales receipt uses "email ="

This is a minor inconsistency.  I can offer a PR, but maybe the designers could weigh in on which method would be preferred?

vanboom:
#375


ruckus:
Thank you!

PR #380, opened 2017-06-02
Adding journal enteries and spec to batch request and response.

nathan-mots:
### What does this do 
Allows sending Journal Entry objects via the batch request object.


ruckus:
Thank you!!

PR #387, opened 2017-07-07
Added purchase order to the list of batch request/response entities. …

jimmybaker:
…Sorted entities alphabetically.


jimmybaker:
Sorry for what looks like a mess. It's because I sorted the lines alphabetically after adding the purchase order entity to the batches.

ruckus:
Thank you!

PR #394, opened 2017-08-31
Exposing Memo field under PurchaseOrder model

ravemz:
It exists in the Quickbooks XML already, adding it to the model to be able to set/get it.


ruckus:
Thank you!

PR #398, opened 2017-10-18
Add support for Query/QueryResponse in Batch operations

### LINKED ISSUE #289: Batch query request
I am trying to use `Quickbooks::Model::BatchRequest` to run several queries in one request.

for example: 

``` ruby
   batch_req = Quickbooks::Model::BatchRequest.new
   batch_req.add('bId1', "Select count(*) from Payments Where CustomerRef = '22'", "query")
   batch_req.add('bId2', "Select count(*) from Payments Where CustomerRef = '33'", "query")
   batch_req.add('bId3', "Select count(*) from Payments Where CustomerRef = '44'", "query")
   # and so on...

   batch_service = Quickbooks::Service::Batch.new
   batch_response = batch_service.make_request(batch_req)
   batch_response.response_items.each do |result|
      puts result.inspect
   end
```

But I am getting error `uninitialized constant String::XML_NODE` caused by `bir.send("#{batch_item.class::XML_NODE.underscore}=", batch_item)` as `batch_item.class` is a String (line 37 in `Quickbooks::Model::BatchRequest`).

In the documentation I found at https://developer.intuit.com/docs/api/accounting/batch I see an example 

``` json
   {
        "bId": "bid4",
        "Query": "select * from estimate"
    }
```

and `add` method generates something like (as I understood)

``` json
{
        "bId": "bid3",
        "operation": "create",
        "Invoice": {
            "Line": [{
                "Amount": 125.00,
                "DetailType": "SalesItemLineDetail",
                "SalesItemLineDetail": {
                    "ItemRef": {
                        "value": "23",
                        "name": "Design"
                    }
                }
            }],
            "CustomerRef": {
                "value": "2"
            }
        }
    }
```

I could not see other method than `add` to create batch query requests. Am I wrong or this is not implemented? 

I would like to have this feature as making a request for each query will take a long time to execute.

jozefvaclavik:
So I needed ability to query in batches. I'm not 100% sure about approach, so any feedback is welcomed.

Fix #289


lib/quickbooks/model/batch_request.rb
```diff
@@ -23,6 +23,7 @@ class BatchItemRequest
     xml_accessor :sales_receipt, from: "SalesReceipt", as: Quickbooks::Model::SalesReceipt
     xml_accessor :time_activity, from: "TimeActivity", as: Quickbooks::Model::TimeActivity
     xml_accessor :vendor, from: "Vendor", as: Quickbooks::Model::Vendor
+    xml_accessor :query, from: "Query"
```
jozefvaclavik: I believe here we could use `xml_accessor :query, from: "Query", as: Quickbooks::Model::QueryResponse` and then below we would just need to `bir.operation = operation unless operation == "query"`.

lib/quickbooks/model/query_response.rb
```diff
@@ -0,0 +1,29 @@
+module Quickbooks
+  module Model
+    class QueryResponse < BaseModel
+      XML_COLLECTION_NODE = "Query"
+      XML_NODE = "Query"
+      REST_RESOURCE = 'query'
```
jozefvaclavik: I think here we need just `XML_NODE`

lib/quickbooks/model/query_response.rb
```diff
@@ -0,0 +1,29 @@
...
+      xml_accessor :invoices, from: "Invoice", as: [Quickbooks::Model::Invoice]
+      xml_accessor :items, from: "Item", as: [Quickbooks::Model::Item]
+      xml_accessor :journal_entries, from: "JournalEntry", as: [Quickbooks::Model::JournalEntry]
+      xml_accessor :payments, from: "Payment", as: [Quickbooks::Model::Payment]
+      xml_accessor :purchase_orders, from: "PurchaseOrder", as: [Quickbooks::Model::PurchaseOrder]
+      xml_accessor :purchases, from: "Purchase", as: [Quickbooks::Model::Purchase]
+      xml_accessor :refund_receipts, from: "RefundReceipt", as: [Quickbooks::Model::RefundReceipt]
+      xml_accessor :sales_receipts, from: "SalesReceipt", as: [Quickbooks::Model::SalesReceipt]
+      xml_accessor :time_activities, from: "TimeActivity", as: [Quickbooks::Model::TimeActivity]
+      xml_accessor :vendors, from: "Vendor", as: [Quickbooks::Model::Vendor]
```
jozefvaclavik: I'm not sure what should be the list of all requests that this gem supports, but I went with the same list that `BatchRequest` supports.

jozefvaclavik:
Im closing this. If anyone is interested in it, please drop a note and reopen it.

Issue #389, opened 2017-07-20
Implement OAuth 2.0


Pro777:
Per this page: https://developer.intuit.com/docs/0100_quickbooks_online/0100_essentials/000500_authentication_and_authorization/connect_from_within_your_app

> As of July 17, 2017, new or existing developers with no previous apps must implement OAuth 2.0. Click here for OAuth 1.0a documentation, available as a reference for existing applications.

ruckus:
Interesting. It appears as if the current `oauth` gem we're using only supports 1.0.

Presumably we'd need to swap that library out for one that is full OAuth 2.0, perhaps:

https://github.com/intridea/oauth2

Methinks this is going to be a non-trivial swap

bradleybennett:
Yeah, I just tried to use your gem for an app I am looking to write to connect Quickbooks Online to Shippo, but couldn't get it to auth and then realized that Intuit is requiring Oauth2 now. Doesn't that mean that anyone using this gem is broken until Oauth2 is integrated? Any plans to tackle this?

ruckus:
I have started to look into this and what its going to take to support OAuth 2.0. I dont have any idea of when I will be able to deliver something, but I am hoping its within 2-3 weeks

Pro777:
@bradleybennett We chatted with Quickbooks support and they said that existing apps can continue using OAuth 1.0 for a while. We just got ours out into the wild last week, so they grandfathered us in.

PatrickatPaperlessPCS:
caught it purgatory myself - any workaround suggestions?

ruckus:
@PatrickatPaperlessPCS I've started working on this and actually have a working proof of concept of completing the OAuth 2 authorization process with making a sample dummy request against the API 

See here:

https://github.com/ruckus/quickbooks-ruby/blob/389-oauth2/webapp/app.rb

You might need to adjust Line 29 with a different Customer ID to fetch - this PoC assumes you have a Customer with ID=1

Depending on your own needs (e.g. you just need something simple on your end to get your own project going, e.g. fetch a list of customers or something) you could basically start rolling your own mini-library, just taking the new OAuth2 connection logic and making your own HTTP calls with an Access Token

The next challenge on my end is making this library compatible with both versions of OAuth. We cannot just rip out support for OAuth1 wholesale and move to OAuth2 -- so all of this has to be done in such a way where the library allows you to set a oauth version to use and the internals adjust accordingly.

Unfortunately I dont have a timeframe for this.

[...]

PR #399, opened 2017-10-20
Don't rebuild locked connection

gregawoods:
Address issue mentioned in https://github.com/ruckus/quickbooks-ruby/issues/389#issuecomment-338240420


ruckus:
Awesome -- thank you!

PR #400, opened 2017-10-23
Raising on invalid OAuth Access Token object type

fabrouy:
I created this PR to prevent a wrong or missing access token initialization.

If for any reason `@oauth` was not an instance of `OAuth::AccessToken` or `OAuth2::AccessToken` it would just not fire the request, keep going and failing on the response parsing without any clue of what went actually wrong.


ruckus:
Awesome, thank you!

PR #401, opened 2017-10-27
Added exists? method to service crud module

fabrouy:
Added method to check for resource existence. Returns `true` or `false`.

This could probably get merged to not only `oauth2` branch, but did not tested there.


ruckus:
Thanks!

Issue #386, opened 2017-07-07
ROXML is unmaintained


plicjo:
This gem is dependent on [ROXML](https://github.com/Empact/roxml), but it hasn't been updated in 5 years.

Ruby 2.4.1 is raising a deprecation warning from inside of the Roxml gem.

```
lib/roxml/definition.rb:156: warning: constant ::Fixnum is deprecated
```

Could the dependency be changed to maintained gems like [Ox](https://github.com/ohler55/ox)?

ruckus:
Its a good question - ROXML is getting long in the teeth.

Perhaps a more long-term approach is to phase out the XML version of this gem in favor of a JSON version which QBO is slowly moving toward anyways.

> Could the dependency be changed to maintained gems like Ox?

Perhaps, but from what I can tell Ox is a generic reader/writer of XML, whereas the use of ROXML is for the conversion of XML to Ruby objects - its a mapping library more than the actual XML parser.

I think the best move is to rewrite this gem using the Representable gem as the workhorse for the XML (and JSON!) serialization:

https://github.com/trailblazer/representable

s-mansouri:
So how long that is take to remove the ROXML? It's great to remove this gem early

ruckus:
@s-mansouri I dont know; I havent had a chance to start researching that task. And I dont know if I will get a chance any time soon. Want to take a crack at it?

pusewicz:
It has just been updated to [version 4.0.0](https://github.com/Empact/roxml/blob/9c6e27c621ed8ff8f030c80fb30ad23333717045/History.txt)

plicjo:
Yeah, it looks like they cleaned house.

PR #410, opened 2017-12-11
Allow roxml 4.0.0

pusewicz:
@ruckus Updates dependencies so that `roxml` 4.0 can be used, fixing warnings in new Ruby versions.

```
roxml-3.3.1/lib/roxml/definition.rb:156: warning: constant ::Fixnum is deprecated
```

Ref #386

### DIFF
--- a/quickbooks-ruby.gemspec
+++ b/quickbooks-ruby.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |gem|
   gem.files = Dir['lib/**/*']
 
   gem.add_dependency 'oauth', '0.4.7'
-  gem.add_dependency 'roxml', '3.3.1'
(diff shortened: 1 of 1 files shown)


ruckus:
Thank you!

evanlouden:
Awesome! Any chance you could cut a release? @ruckus

ruckus:
@evanlouden good idea. Current version is `0.5.1` - I'm not up on semver, would this suggest a `0.6.0` release, or a `0.5.2` one?

evanlouden:
@ruckus I'm pretty new to development but guess is it would be a MINOR verision `0.6.0` because it adds functionality in a backwards-compatible manner but isn't a bug fix (PATCH 0.5.2).

Thanks!

ruckus:
`0.6.0` pushed to RubyGems!

PR #412, opened 2017-12-19
Fix Content-Type in Send Invoice

gabrielbg:
According to documentation in https://developer.intuit.com/docs/api/accounting/invoice `Content-Type` for sending Invoice email must be `application/octet-stream`

This PR intends to fix this by sending the correct `Content-Type` in the service call to `do_http_post`

The error I was originally getting was ```NoMethodError: undefined method `bytesize' for #<Hash...```, and after fixing this I started getting `400` status due to the `Content-Type`


ruckus:
Thank you!

PR #416, opened 2018-02-19
Add support to follow 302 redirects in do_http

srushe:
Recently I have been receiving occasional 302 responses from the QuickBooks API and thus discovered that the gem doesn't follow the redirects. This PR updates `do_http` method in the base service to follow 302s for both GET and POST requests, as well as adding tests for said method. It also updates `stub_request` to support headers being provided, and updates all uses of it to handle that.

As part of this PR I've also updated `rspec` from 2.13 to 2.14.1. I needed `have_received` for the specs, *and* `2.14.1` is the most up-to-date version before deprecation notices appear regarding version 3.

### DIFF
--- a/lib/quickbooks/service/base_service.rb
+++ b/lib/quickbooks/service/base_service.rb
@@ -239,7 +239,12 @@ module Quickbooks
           else
             raise "Do not know how to perform that HTTP operation"
           end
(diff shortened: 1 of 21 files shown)


ruckus:
Thank You so much!

PR #418, opened 2018-02-26
Add exception for unhandled redirect instead of raising a RuntimeError

stevedev:
Having to catch a runtime error and inspect its message to match against an unhandled redirect is causing some pain for us (and there have been a lot of unhandled HTTP redirect errors today). This could be a breaking change for some people but I thought I'd put it out there as an idea to see if it's worth implementing into the gem.

### DIFF
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -226,6 +226,7 @@ module Quickbooks
   class TooManyRequests < StandardError; end
   class ServiceUnavailable < StandardError; end
   class MissingRealmError < StandardError; end
+  class UnhandledHttpRedirect < StandardError; end
 
   class IntuitRequestException < StandardError
     attr_accessor :message, :code, :detail, :type, :request_xml, :request_json
--- a/lib/quickbooks/service/base_service.rb
+++ b/lib/quickbooks/service/base_service.rb
@@ -269,7 +269,7 @@ module Quickbooks
             response
           end
         when 302
-          raise "Unhandled HTTP Redirect"
+          raise Quickbooks::UnhandledHttpRedirect
         when 401
           raise Quickbooks::AuthorizationFailure
         when 403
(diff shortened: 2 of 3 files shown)


ruckus:
Good call. I can see this being useful. And yes, inspecting exception string messages is not the cleanest.

In terms of redirects - would this recent PR address your issues?

https://github.com/ruckus/quickbooks-ruby/commit/50e9dec051bdecb7da553de9982e1726d62d4cae

stevedev:
I was looking at that yesterday too and I think it will be good as we will then be handling some of the 302s. There's still the case of the unhandled redirects. The problem we were seeing yesterday (and it happens fairly often) is that the QBO API was redirecting because it was down instead of throwing a 503 error. I'm super curious what will happen in that case... 

unhandled http redirects - 30 days

lib/quickbooks-ruby.rb
```diff
@@ -226,6 +226,7 @@ class ThrottleExceeded < Forbidden; end
   class TooManyRequests < StandardError; end
   class ServiceUnavailable < StandardError; end
   class MissingRealmError < StandardError; end
+  class UnhandledHttpRedirect < StandardError; end
```
drewish: Not specific to this PR but I'm wondering what @ruckus would think about introducing a base QBO error that all of these can inherit from so there's an single one to target for a rescue block, e.g:
```
class Error < StandardError; end
class InvalidModelException < Error; end
class AuthorizationFailure < Error; end
...
```
Then you can just catch all of these with:
```
rescue Quickbooks::Error => e
```
```

ruckus: @drewish thats probably a good idea. A PR would be much appreciated. Thank you!

drewish: @ruckus Great, I opened up https://github.com/ruckus/quickbooks-ruby/pull/445

drewish: That was merged so this PR will need a rebase now.

PR #431, opened 2018-05-24
Treat http 502 as ServiceUnavailable instead of `RuntimeError`

serixscorpio:
We have come across http 502 (Bad Gateway), and think it might be sensible to treat http 502 as a `Quickbooks::ServiceUnavailable` instead of generic `RuntimeError` with msg "HTTP Error Code: 502 ...".  Not sure if this will be a breaking change for some people, but perhaps it's sufficiently similar to http 504 in terms of error handling.


ruckus:
Thank you!

PR #435, opened 2018-07-13
Fixes #434

### LINKED ISSUE #434: Reports with columns string columns
Hi,

Quickbooks::Model::Reports assumes each reports has **one** column with text and the rest are number.

Reports with multiple text columns will fail ie
```
ArgumentError: invalid value for BigDecimal(): "Invoice"
from /Users/****/.rvm/gems/ruby-2.5.0/gems/quickbooks-ruby-0.6.2/lib/quickbooks/model/report.rb:45:in `BigDecimal'
```

Proposed solution: Evaluate the value of the column to determine the result. Blank values - `nil`, Numeric - `BigDecimal`, otherwise `String`. This will ensure not change to existing reports but will allow text to be other columns.

davidrichey:
This will allow reports with multiple strings to render [Example Row](https://github.com/ruckus/quickbooks-ruby/pull/435/files#diff-7ec509edffa99799ffbe734001960ed8R52)

Regex:
^\d+$ - `5`
^\d+.\d+$ - `5.5`
^-\d+ - `-500`
^-\d+.\d+$ - `-5.5`
^.\d+$ - `.55`

### DIFF
--- a/lib/quickbooks/model/report.rb
+++ b/lib/quickbooks/model/report.rb
@@ -37,13 +37,9 @@ module Quickbooks
         row_node.elements.map.with_index do |el, i|
           value = el.attr('value')
 
-          if i.zero? # Return the first column as a string, its the label.
-            value
-          elsif value.blank?
-            nil
-          else
-            BigDecimal(value)
-          end
+          next nil if value.blank?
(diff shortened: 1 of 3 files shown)


ruckus:
Thank you!

PR #436, opened 2018-07-23
Validate TimeActivity description length

john-goldsmith:
* Fix validation error typo (`EmployeeRef` was referred to as `VendorRef`)
* Add max length validation to TimeActivity description
  * Per [the TimeActivity docs](https://developer.intuit.com/docs/api/accounting/timeactivity#/id-timeactivityresponse), the maximum character length of the `Description` property is 4000
  * When invalid, the QuickBooks API will respond with:

```js
{
  "Fault": {
    "Error": [
      {
        "Message": "String length is either shorter or longer than supported by specification",
        "Detail": "String length specified does not match the supported length. Min:0 Max:4,000 supported. Supplied length: 4,006",
        "code": "2050",
        "element": "Description"
      }
    ]
  }
}
```

All specs still pass (although I did get warnings about `Fixnum` being deprecated using Ruby 2.5.0):

```
418 examples, 0 failures
3014 / 3229 LOC (93.34%) covered
```


ruckus:
Thank you!

PR #440, opened 2018-09-12
Add support for Minor Version 1 which adds TaxInclusiveAmt attribute …

SpiffyStores:
…Line.SalesItemLineDetail, Line.ItemBasedExpenseLineDetail, and Line.AccountBasedExpenseLineDetail line types.

This adds support for additional attributes used outside of the US that allow prices to be displayed as tax inclusive amounts.

### DIFF
--- a/lib/quickbooks/model/invoice.rb
+++ b/lib/quickbooks/model/invoice.rb
@@ -14,6 +14,7 @@ module Quickbooks
 
       #== Constants
       REST_RESOURCE = 'invoice'
+      MINORVERSION = 1
       XML_COLLECTION_NODE = "Invoice"
       XML_NODE = "Invoice"
       EMAIL_STATUS_NEED_TO_SEND = 'NeedToSend'
--- a/lib/quickbooks/model/sales_item_line_detail.rb
+++ b/lib/quickbooks/model/sales_item_line_detail.rb
@@ -9,6 +9,7 @@ module Quickbooks
       xml_accessor :quantity, :from => 'Qty', :as => BigDecimal, :to_xml => Proc.new { |val| val.to_f }
       xml_accessor :tax_code_ref, :from => 'TaxCodeRef', :as => BaseReference
       xml_accessor :service_date, :from => 'ServiceDate', :as => Date
+      xml_accessor :tax_inclusive_amount, :from => 'TaxInclusiveAmt', :as => BigDecimal, :to_xml => Proc.new { |val| val.to_f }
 
       reference_setters :item_ref, :class_ref, :price_level_ref, :tax_code_ref
     end
--- a/lib/quickbooks/service/invoice.rb
+++ b/lib/quickbooks/service/invoice.rb
@@ -6,6 +6,21 @@ module Quickbooks
         delete_by_query_string(invoice)
       end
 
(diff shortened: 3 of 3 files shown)


SpiffyStores:
Missed some code and updates to 0.6.2

lib/quickbooks/service/base_service.rb
```diff
@@ -289,7 +289,7 @@ def check_response(response, options = {})
         when 429
           message = parse_intuit_error[:message]
           raise Quickbooks::TooManyRequests, message
-        when 502, 503, 504
```
ruckus: Why the removal of the 502 code here?

lib/quickbooks/version.rb
```diff
@@ -1,5 +1,5 @@
 module Quickbooks
 
-  VERSION = "0.6.1"
+  VERSION = "0.6.2"
```
ruckus: No need to bump this yourself. I will do it pre-release

ruckus:
Thanks for the PR. The specs are failing. We'll need a green build for a merge. Thanks again!

lib/quickbooks/service/base_service.rb
SpiffyStores: These updates came from the 0.6.2 version of the gem. I merged them in from the gem as I was surprised that they were not in the github repository. Are you the only owner of the gem, as it seems to have been updated independently of the github repository?

ruckus: Ah gotcha. Yep, my bad. After cutting `0.6.2` back in May I forgot to push my change to `version.rb`; which I just did. However, it was only `version.rb` and `HISTORY.md` that didnt get pushed. All other files (such as `lib/quickbooks/service/base_service.rb` and the addition of the `502` status check have been in the repo since May)

SpiffyStores: These have now been reverted and can be ignored. I was confused by the status of the gem as it didn't match the repo. Do I need to do anything more for this pull request?

mnaser:
@SpiffyStores 

Do you plan on fixing those changes?  If not, I can take those changes and create a new PR out of them?

SpiffyStores:
> 
> 
> @SpiffyStores
> 
> Do you plan on fixing those changes? If not, I can take those changes and create a new PR out of them?

There's nothing wrong with the changes. The problem in the Travis CI build seems to have nothing to do with the updates, and is reporting that there are no environment variables set. This is not something I can fix.

Any advice?

SpiffyStores:
> 
> 
> > @SpiffyStores
> > Do you plan on fixing those changes? If not, I can take those changes and create a new PR out of them?
> 
> There's nothing wrong with the changes. The problem in the Travis CI build seems to have nothing to do with the updates, and is reporting that there are no environment variables set. This is not something I can fix.
> 
> Any advice?

Ok, I can see that the specs need updating to support the additional minorversion parameter, but at the moment, I don't have much of an understanding on how the specs have been written and how they interact with QB.

I know that the actual code updates are good, but I may need some help and time to get the spec code updated to match the code update.

vinayakkatkar:
@ruckus any idea on when can be this SalesItemLineDetail will have TaxInclusiveAmt attribute?

PR #445, opened 2018-11-29
Add a base class to easily rescue all QBO exceptions

drewish:
This is a follow up to something that came up on https://github.com/ruckus/quickbooks-ruby/pull/418/files#r234803100 

By having all the errors inherit from a single class we can then just rescue `Quickbooks::Error` to handle all of this gem's exceptions.


ruckus:
Thank you!

Issue #329, opened 2016-09-02
Upgrading OAuth


drewish:
I'm curious if there's a reason you're not using the 0.5.x release of the oauth gem. I'd like to use a newer version to get access to the rest-client proxy.

ruckus:
Good question. I dont remember the exact details but I do remember it that version of the OAuth gem playing nicely with multipart uploads - which we do here:

https://github.com/ruckus/quickbooks-ruby/blob/master/lib/quickbooks/util/multipart.rb

At the time of developing the multipart stuff I had tried later version of the gem and I was running into issues. 

But that was awhile ago and could very well be resolved. Its probably worth looking into:

1. Updating to latest OAuth gem version
2. Attempting an Attachable upload

If it works then we potentially no reason why we cannot upgrade across the board.

drewish:
Thanks, I'll spend a little time testing that in the next week or two. I need to wrap up some other tasks before I get back to this.

drewish:
I'm going to close this. The new OAuth stuff seems pretty buggy and I don't want to spend the time debugging it.

PR #449, opened 2019-01-03
Relax oauth gem dependency

jnraine:
The current version constraint requires every app using this gem to also use `oauth` version 0.4.7. For apps that already use `oauth`, as is the case with Clio, bundler fails to resolve a version of `oauth` that works satisfies all conditions.

In this commit, I’ve relaxed the version constraint to allow anything >= 0.4.5 but < 0.5. To get around this constraint, we’ve been using a [fork](https://github.com/nathan-mots/quickbooks-ruby) but I’d love to get back to using the proper gem.

Is this a change that you'd be open to making?

### DIFF
--- a/quickbooks-ruby.gemspec
+++ b/quickbooks-ruby.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |gem|
 
   gem.files = Dir['lib/**/*']
 
-  gem.add_dependency 'oauth', '0.4.7'
+  gem.add_dependency 'oauth', '~> 0.4.5'
   gem.add_dependency 'roxml', '>= 3.3.1', '< 4.1'
   gem.add_dependency 'nokogiri'  # promiscuous mode
   gem.add_dependency 'activemodel' # promiscuous mode


ruckus:
Yeah this sounds reasonable. I do remember pinning it to that version because there was an issue with multipart uploads for attachments. That was a long time ago and I don't remember all of the details.

ruckus:
Thank you!

jnraine:
@ruckus awesome! Thanks for the quick response. I'll point our app at `ruckus/quickbooks-ruby` master and then switch to the latest version of the gem once you cut a release.

When do you think the next release will go out?

drewish:
I’m not a maintainer but worth noting that you can always point your Gemfile at the master repo’s branch.

jnraine:
@drewish we’re on the same wavelength 🌊—I opened that PR this afternoon. 😄✌🏻

ruckus:
@jnraine I just pushed `0.6.5` to RubyGems with your update. thanks!

orangethunder:
Has anyone attempted to use the gem with oauth >= 0.5.0?

Looks like 0.4.7 was released in 2012, and 0.5. in 2016, but I don't see any breaking changes in the ouath HISTORY file. If no-one has info on this, I might attempt forking this to see if it works.

drewish:
@orangethunder give https://github.com/ruckus/quickbooks-ruby/issues/329 a look

jnraine:
@orangethunder If you fork and try with a newer version of `oauth`, I'd be curious to hear how it goes. We'll need to upgrade our version of `oauth` in the not-too-distant future too.

PR #452, opened 2019-01-07
adding LinkedTxn collection into Quickbooks::Model::Bill per v3 API spec

thaiden:
## Intent

To un parse `LinkedTxn` attribute into `QuickBooks::Model::Bill` from XML per (API description)[ https://developer.intuit.com/app/developer/qbo/docs/api/accounting/most-commonly-used/bill]

Example of XML payload
```xml
<Bill xmlns="http://schema.intuit.com/finance/v3">
  .....
  <LinkedTxn>
    <TxnId>3527</TxnId>
    <TxnType>BillPaymentCheck</TxnType>
  </LinkedTxn>
   .....
</Bill>
```

## Usage:
```ruby
bill.linked_transactions.each do |trx|
  # some logic
end
```

### DIFF
--- a/lib/quickbooks/model/bill.rb
+++ b/lib/quickbooks/model/bill.rb
@@ -19,6 +19,8 @@ module Quickbooks
 
       xml_accessor :private_note, :from => 'PrivateNote'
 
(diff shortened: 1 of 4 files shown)


ruckus:
Thank you!

Issue #451, opened 2019-01-05
Is there a way to assign multiple attributes after object been initialized?


wayne5540:
Hi, is there a way to assign multiple attributes into model after initialized? For example:

```rb
incoive = Quickbooks::Model::Invoice.new
invoice.assign_attributes(attributes)
```
In this way, we don't need to manually assign attribute by using setter respectively.

I think it's useful and nice to have, in ActiveModel there is a module called [AttributeAssignment](https://api.rubyonrails.org/classes/ActiveModel/AttributeAssignment.html), maybe we can considering adding it into BaseModel since we already have ActiveModel dependency?

wayne5540:
My pleasure, PR: https://github.com/ruckus/quickbooks-ruby/pull/453

PR #453, opened 2019-01-08
BaseModel includes ActiveModel::AttributeAssignment

wayne5540:
## Changes

Add following functions by including ActiveModel::AttributeAssignment:

* Quickbooks::Model::BaseModel#assign_attributes (Added test)
* Quickbooks::Model::BaseModel#attributes= (No test)

## Notes

* Added a `assign_attributes` test case, it's duplicated with ActiveModel's test but I think it's ok to duplicate it so we can make sure it works on this gem.
* While working on this gem, I found the rspec dependency is still `2.14.1` which is relatively old compare to current version `3.8.0`, this bothered me because I was trying to use `aggregate_failures` function to test getters from `#assign_attributes` separately so there won't be dependency with `#attributes` test, are you considering update `rspec` to current version in the future? Happy to help with it as well.

## Reference

* https://github.com/ruckus/quickbooks-ruby/issues/451
* https://api.rubyonrails.org/classes/ActiveModel/AttributeAssignment.html

### DIFF
--- a/lib/quickbooks/model/base_model.rb
+++ b/lib/quickbooks/model/base_model.rb
@@ -4,6 +4,7 @@ module Quickbooks
       include Definition
       include ActiveModel::AttributeMethods
       include ActiveModel::Validations
+      include ActiveModel::AttributeAssignment
       include Validator
       include ROXML
 
--- a/spec/lib/quickbooks/model/base_model_spec.rb
+++ b/spec/lib/quickbooks/model/base_model_spec.rb
@@ -74,7 +74,7 @@ describe "Quickbooks::Model::BaseModel" do
     it "assigns values into attributes" do
       foo_model.assign_attributes({ baz: "quiz", bar: { foo: 1 }, amount: 30 })
 
-      foo_model.attributes.should eq("baz" => "quiz", "bar" => {"foo" => 1}, "amount" => 30)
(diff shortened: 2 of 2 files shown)


wayne5540:
Test breaks because `ActiveSupport` requires Ruby > 2.2.2.

drewish:
Might be worth rebasing this PR against master to see if the tests will be happier with all the gem bumps that have happened since then.

wayne5540:
I see the reason why this PR failed the test. Because `ActiveModel::AttributeAssignment` was introduced by rails 5+. [reference](https://github.com/rails/rails/commit/2606fb339797a99c50e531105fc92071cef3db01)

Run the CI with Ruby verison < 2.3 will use activemode 4.2.11 version which has not yet implemented `ActiveModel::AttributeAssignment` module.

To make this PR works we need to lock dependency to `activemodel > 5.0` which I don't know if it worth.

ruckus:
Thanks for investigating @wayne5540 

No we definitely cannot lock to `> 5.0`. I'm still on `4.x` (just upgraded my massive legacy app from rails3 last week!)

wayne5540:
@ruckus Yes, agree! Feel free to close this PR.

ruckus:
Closing for now. This PR requires `activemodel > 5.0`

PR #456, opened 2019-01-14
log response headers.   Most interesting is the Intuit TID, which Intuit Support may ask for.

bretweinraub:
We may just want the Intuit TID.  If so, instead of "each_header", we can just log

  response['intuit_tid']


ruckus:
Probably sane to include it all. The user has enabled verbose logging.

ruckus:
Thank you!

bretweinraub:
Thanks for the very useful gem!

PR #457, opened 2019-01-15
Update HISTORY.md

thaiden:
updating, since you've pointed to wrong Vitalijs


ruckus:
Thank you! Sorry about that.

PR #459, opened 2019-01-24
Allow to use any patch version of Oauth2 v1.4 gem

senzpo:
Oauth2 library aims to adhere to Semantic Versioning 2.0.0, it means that we can use any patch version. Is it possible to confirm this pull? I want to use oauth2 1.4.1 because it is impossible to use 1.4.0 in my application. I already checked 1.4.1 in my dev env.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,11 @@ rvm:
   - 2.0.0
   - 2.1.0
   - 2.2.0
+  - 2.3.0
+  - 2.4.0
+  - 2.5.0
 # Bundler 1.13 added support for `ruby RUBY_VERSION` in the Gemfile
-before_install: gem install bundler -v '>= 1.13.0'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,8 +4,11 @@ rvm:
   - 2.0.0
   - 2.1.0
   - 2.2.0
+  - 2.3.0
+  - 2.4.0
+  - 2.5.0
 # Bundler 1.13 added support for `ruby RUBY_VERSION` in the Gemfile
-before_install: gem install bundler -v '>= 1.13.0'
+before_install: gem install bundler -v '1.17.3'
```
senzpo: I set this version because it's the last version to allow use ruby v.1.9.3 Without this fix it's possible that Travis will use bundler wich not allowed ruby less than 2.3.0

ruckus:
makes sense; thank you!

PR #468, opened 2019-04-23
exposing time attribute from response received via CDC endpoint

thaiden:
# Why
This value can be useful when running various sync and want to rely on previous sync time before executing a new one.

Pretty small change but its cleaner than forcing responsibility to lookup it in `xml.root.attribute`

### DIFF
--- a/lib/quickbooks/model/change_data_capture.rb
+++ b/lib/quickbooks/model/change_data_capture.rb
@@ -19,6 +19,13 @@ module Quickbooks
          data
       end
 
+      # time when refresh was requests from cdc/ endpoiint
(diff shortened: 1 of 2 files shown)


ruckus:
Thanks, makes sense. Looks like you're using the Safe Navigation operator introduced in Ruby 2.3.

Yes yes I know we all should be not running such old versions of Ruby - but the world is not so perfect. Would you be able to refactor this into a more backwards-compatible approach?

Thank you!

thaiden:
@ruckus sure, my bad should've paid attention to minimum ruby version.

thaiden:
@ruckus addressed your feedback

ruckus:
Thank you!

PR #469, opened 2019-04-25
Closes #467 - Allow disconnect of OAuth2 tokens

### LINKED ISSUE #467: Support for OAuth2 disconnect/revoke
Hi! We've been using this gem with much success, so thank you all for your work on it! 

We're in the process of migrating to OAuth 2 for authentication to QuickBooks Online, and we've hit a snag on the`389-oauth2` branch. There doesn't seem to be a way to revoke or disconnect the access token. 

Our OAuth 1.0 code simply called `Quickbooks::Service::AccessToken#disconnect` and all was well. Authenticating with OAuth 2 and attempting to disconnect using the same method yields an error from Intuit:

```xml
<?xml version="1.0" encoding="utf-8"?>
<PlatformResponse xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://platform.intuit.com/api/v1">
  <ErrorMessage>Authentication required</ErrorMessage>  
  <ErrorCode>22</ErrorCode>
  <ServerTime>2019-04-22T19:05:08.3346136Z</ServerTime>
</PlatformResponse>
```

I'm assuming this is because we're hitting an OAuth 1.0 service and trying to disconnect our OAuth 2 token from it, and that service doesn't see our user as authenticated. Is that correct?

Are there currently any methods to disconnect or revoke an OAuth 2 token implemented? Or perhaps I am just doing something wrong, or using the wrong service to disconnect my token?

jaredmoody:
- OAuth2 disconnect requires POST instead of GET
- We can't use `do_http_post` becuase it uses `OAuth2::AccessToken#post` which overwrides the `Authorization` header, so use Faraday to send the `POST` request to disconnect

### DIFF
--- a/lib/quickbooks/service/access_token.rb
+++ b/lib/quickbooks/service/access_token.rb
@@ -24,14 +24,14 @@ module Quickbooks
       def disconnect
         result = nil
 
-        disconnect_url = nil
-
         if oauth_v1?
-          disconnect_url = DISCONNECT_URL_OAUTH1
+          response = do_http_get(DISCONNECT_URL_OAUTH1)
         elsif oauth_v2?
-          disconnect_url = DISCONNECT_URL_OAUTH2
+          conn = Faraday.new
+          conn.basic_auth oauth.client.id, oauth.client.secret
+          raw_response = conn.post(DISCONNECT_URL_OAUTH2, { token: oauth.refresh_token || oauth.token })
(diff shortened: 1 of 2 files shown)


ruckus:
Thank you!

PR #470, opened 2019-04-25
#467 - Fix error when OAuth2 disconnect succeeds

jaredmoody:
Even when setting the accepts header to xml, the OAuth2 disconnect endpoint always returns an empty body.

So, I updated the stubbed requests in the specs to duplicate the bug, and changed the approach to construct a `Quickbooks::Model::AccessTokenResponse` with the appropriate fields for success/fail.

I don't love this solution, but short of modifying `Quickbooks::Model::AccessTokenResponse` to function completely differently, this seems ok - the external interface remains the same.

/cc @Zajn


ruckus:
Thank you!!!

PR #478, opened 2019-08-08
Feature/group line detail

### LINKED ISSUE #464: Bundled Items in SalesReceipt Creation
Is there a way to create a sales receipt containing a bundled item?

I can achieve this when creating an invoice, but I get the error `NoMethodError - undefined method group_line_detail!' for #<Quickbooks::Model::Line:0x0055abebb97598>`  when attempting the same with a sales receipt. That makes sense as there isn't a group line detail option in the Line model.

Am I missing something or is this functionality not yet available?

gorandev:
Fixes #464


ruckus:
Thank you!

gorandev:
Thank *you* for all your work!

PR #479, opened 2019-08-19
Invoice Link added

railsfactory-palpandi:
Added Invoice Link option read only attribute in invoice model
minorversion changed from 21 to 37


ruckus:
Thank you!

PR #481, opened 2019-09-23
OAuth 2 support (merge conflicts with master resolved)

yukideluxe:
Hello!

We are using `quickbooks-ruby` and we are about to release OAuth v2 support! I took the freedom to merge the latest master with `389-oauth2` and I'm creating this PR just in case it helps you to push the OAuth2 support forward!

Feel free to close it if you want to merge it yourself @ruckus! Thanks for all your work on this <3


lib/quickbooks/model/base_model.rb
```diff
@@ -2,6 +2,7 @@ module Quickbooks
   module Model
     class BaseModel
       include Definition
+      include ActiveModel::AttributeMethods
```
yukideluxe: I had to add this because the tests were broken 🤔 There's a PR about that: https://github.com/ruckus/quickbooks-ruby/pull/453/files

ruckus:
Thanks! I think I just need to review the README and update it for OAuth2 stuff and then I say we can merge in this PR. 

Do you have any README notes that you think might be good to include? From what I can remember it should just be:

* New Authentication setup
* Token Refreshing

pjg:
> Token Refreshing

This would probably be best split into two parts: `access_token` refresh and `refresh_token` refresh as those two are distinct.

README.md
```diff
@@ -57,8 +57,99 @@ you need to config the gem to run in sandbox mode:
...
+### Access Token Validity and Token Refresh
+
+Each access token is only valid for one hour. The access token and refresh token can be refreshed directly by using OAuth Client:
+
+```ruby
+new_access_token = access_token.refresh!
+```
+
+The token must be assigned to a variable to prevent the loss of your new access token, which will void your credentials and a new set of credentials have to be acquired by authorizing the application again.
+Unauthorized (expired) access to the API will raise a `Quickbooks::AuthorizationFailure` error.
```
pjg: If I try refreshing with a made up refresh token, the exception I get is `OAuth2::Error`:

```
OAuth2::AccessToken.new(QB_OAUTH2_CONSUMER, access_token, { refresh_token: 'abc' }).refresh!

OAuth2::Error: invalid_grant: Incorrect or invalid refresh token
{"error_description":"Incorrect or invalid refresh token","error":"invalid_grant"}
from /.../.bundle/gems/oauth2-1.4.1/lib/oauth2/client.rb:119:in `request'
```

yukideluxe:
Thanks for the feedback!

> And, fwiw, I can confirm that this branch is working nicely with QuickBooks OAuth v2 app.

@pjg yes! It's already working in our production servers!  😬 

> This should probably mention both access_token and refresh_token refresh as those two are distinct. This info is already in the README, but perhaps could be expanded a little bit more (and mention the 100 days expiration time of refresh_token).

@pjg There's already a link (that I just updated) to the official documentation where people can read more about how the token expiration works. I'd say that's enough! What do you think? https://github.com/harvesthq/quickbooks-ruby/blame/with-oauth2-support/README.md#L140

@ruckus I went through the documentation and fixed some stuff. I took the chance to put some code example to revoke tokens with the gem!

README.md
```diff
@@ -120,15 +121,17 @@ Once you have your user's OAuth token, you can re-use the `OAuth Consumer` and c
...
 
 ### Access Token Validity and Token Refresh
 
 Each access token is only valid for one hour. The access token and refresh token can be refreshed directly by using OAuth Client:
 
 ```ruby
-new_access_token = access_token.refresh!
+redirect_uri = oauth_callback_quickbooks_url
+
+new_access_token = access_token.refresh!(:redirect_uri => redirect_uri)
```
pjg: I can do `access_token.refresh!` just fine without providing the `redirect_uri` param.

yukideluxe: I would have sworn that `redirect_uri` was mandatory in the refresh request, too. I might have dreamt about that... 😂 I fixed it! Thanks

pjg:
I would perhaps add one more line with code to the README in the section about refreshing the token:

```ruby
quickbooks_credentials.update(
  access_token: new_access_token.token,
  refresh_token: new_access_token.refresh_token
)
```

ruckus:
Good points @pjg . 

My memory is rusty and I'm looking into how this would be accomplished by digging into the `BaseService` methods.

If I recall, oauth1 library would not raise exceptions on its own and we checked the HTTP status code, which allowed us to handle the response handling (and raising of errors) ourselves in the `check_response` method.

Looking at the OAuth2 gem it appears that it does raise exceptions itself for a non-200 status code. Meaning that our existing `check_response` method never gets called; thereby implying that we would need to wrap `do_http` in its own begin/rescue block to catch this. 

However, looking a little more at the OAuth2/Client itself:

https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2/client.rb#L119

It appears we _might_ be able to set `raise_errors: false` in the initial client configuration; thereby not raising the error and letting our current `check_response` fall through.

Well, to be more clear, we would suggest people set this config value when they create the intial client; as its not up to us internally.

@pjg do you think this would work?

pjg:
As a developer I value existing conventions/behaviour to be persisted between new releases. As such, I would like to have the `raise_errors` setting be configurable by this gem's user, but I'd opt for having it be set to `false` by default.

ruanltbg:
@ruckus do you have an update on this?

yukideluxe:
Thanks for merging this @ruckus! I see that you've released a new version of the gem already but [shouldn't we update the version file](https://github.com/ruckus/quickbooks-ruby/blob/master/lib/quickbooks/version.rb)? :)

Thanks again!

ruckus:
Thanks @yukideluxe  -- forgot to push. Doh!

lib/quickbooks/service/base_service.rb
```diff
@@ -293,7 +342,7 @@ def check_response(response, options = {})
         when 429
           message = parse_intuit_error[:message]
           raise Quickbooks::TooManyRequests, message
-        when 502, 503, 504
+        when 503, 504
```
nishantdania: Why was the `502` error code removed from this check ?

PR #483, opened 2019-10-02
Hotfix/auto credit apply

mabdullahkhalil:


drewish:
Sorry I hadn't seen your PR but I'd also needed that same field and I'd put in https://github.com/ruckus/quickbooks-ruby/pull/492 which adds that value as well as some others in the SalesFormsPref section. That PR was merged so I think we could close this.

PR #484, opened 2019-10-18
Error code 502 throws ServiceUnavailable

nishantdania:
Error code 502 is being thrown as a StandardError in the new release.

It was a ServiceUnavailable Error prior to the new release as the change was made here >> https://github.com/ruckus/quickbooks-ruby/pull/431

Bringing it back as it seems like a bug.


ruckus:
Thank you; yeah somehow that slipped out.

PR #488, opened 2019-11-05
Add minorversion to account service and update query url

colmheaney:
This PR adds a minorversion of 13 to the Account model and updates `Quickbooks::Service::Account` `url_for_query` method to include that minorversion.

See https://github.com/ruckus/quickbooks-ruby/issues/487 for an explanation of why this is desirable.


ruckus:
Thank you!

PR #490, opened 2019-11-26
Better errors

drewish:
I've been playing around with this to try to get some better errors to help me with some debugging. 

I'd initially wanted the error message carried over to AuthorizationFailures, in there I noticed that we aren't setting the `element` correctly it was always getting the `code`'s value. 

I tried to separate out the logging and response parsing:
- `do_http` now does the request and response logging... we might not want all that in there but at least it's together now.
- `check_response` now calls the `parse_*` methods. It felt odd to me those were a side effect of the `log_response_body` method.
- There's a fair bit of work to carry the `intuit_tid` header out to the `IntuitRequestException`. I'd been asked for that value by Intuit's tech support to diagnose a 500 and, annoyingly had to enable full logging to discover they weren't including any headers in their error response. I'm not sure how useful that would be for anyone else so happy to revert that part if you'd like.

### DIFF
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -231,7 +231,16 @@ module Quickbooks
 
   class Error < StandardError; end
   class InvalidModelException < Error; end
-  class AuthorizationFailure < Error; end
+  class AuthorizationFailure < Error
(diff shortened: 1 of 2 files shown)


ruckus:
Thanks @drewish ; makes sense.

Issue #454, opened 2019-01-08
Considering update rspec dependency to current version?


wayne5540:
We are using `2.14.1` now, the latest version is `3.8.0`, are you considering update it to current version, it will break some tests because of removed syntax like `its` and `be_false`, but they are easy to fix. Happy to help.

Reference: https://github.com/ruckus/quickbooks-ruby/pull/453

drewish:
I like the idea so I opened up https://github.com/ruckus/quickbooks-ruby/pull/494 to start working on this.

PR #494, opened 2019-12-06
Update rspec

drewish:
Looking to address #454 by bumping to the current version of rspec and starting to switch over to the new syntax.


ruckus:
Fantastic, thank you!

drewish:
I think you could close https://github.com/ruckus/quickbooks-ruby/issues/454 now if you wanted.

PR #498, opened 2020-01-02
Fix Content-Type in send sales receipt

mingxiangchan:
Fixes the `undefined method 'bytesize' for #<Hash...` error triggered when
sending a sales receipt.

Searched around and found #412  which implemented the fix for sending invoices, but hadn't been implemented for sales receipt yet.


ruckus:
Thank you!

PR #499, opened 2020-01-13
Finish update to rspec 3

drewish:
Discovered the https://github.com/yujinakayama/transpec gem and decided to run it to finish the upgrade to rspec 3. Wow, I can't believe I did the first part by hand.


ruckus:
Awesome, thank you! Yah that was a lot of manual drudgery. I started chipping away at it as well and it was going to be tedious.

Issue #501, opened 2020-01-23
undefined method `oauth_v1?'


sbrych:
Hi.

There is some left-over code missed by the OAuth v1 purge and it causes said error when calling `#disconnect` on `Quickbooks::Service::AccessToken`

// Edit. Took the liberty of preparing a PR (https://github.com/ruckus/quickbooks-ruby/pull/502)

PR #502, opened 2020-01-23
Remove old oauth_v1 check

sbrych:
# Description
The `#oauth_v1?` method was removed and the `#oauth_v2?` method was commented out. And it causes exceptions when trying to disconnect.

# Fixes
https://github.com/ruckus/quickbooks-ruby/issues/501


ruckus:
Fantastic, thank you so much!

PR #505, opened 2020-02-10
fix fetch_by_id in purchase order service

atemena:
Commit 0c9eb72 broke `fetch_by_id` in the `PurchaseOrder` service. Using the method would result in: `Object Not Found : TxnType does not match read: Purchase Order expected: Invoice`.

This was due to fetch_by_id using the invoice endpoint instead of the purchase order endpoint.


ruckus:
Ugh, my bad. Thank you so much for the PR!

PR #507, opened 2020-02-23
Add callbacks around HTTP requests

anaprimawaty:
Add `before_request`, `around_request` and `after_request` callbacks around HTTP requests. This is useful for instrumentation, logging / metric gathering or controlling rate limits.

Inspiration from https://github.com/waynerobinson/xeroizer/pull/251 and https://github.com/waynerobinson/xeroizer/pull/344


ruckus:
Interesting! Thank you!

PR #511, opened 2020-03-09
Escape regex matching numbers in report parsing

mehwoot:
`.` will match anything, not just the literal character.  This means that the regex that is attempting to filter only numbers will match unexpected strings, such as "T1", which matches `/^.\d+$/`, but does not match `^\.\d+$/`
Previously when values such as "T1" appear in a report, an exception will be raised by BigDecimal because it cannot parse the value


ruckus:
Thank you!

PR #513, opened 2020-03-22
Raise IntuitRequestException for unhandled status code

anaprimawaty:
When there's an unhandled status code, raise `Quickbooks::IntuitRequestException` instead of `RuntimeError` to allow better error handling on client side.


drewish:
What code are you seeing? It would probably be best to just return `Quickbooks::ServiceUnavailable`s for all the 500s. I can’t remember if we added a base class for the 400s.

anaprimawaty:
Ah it was actually 502, but it has been fixed here https://github.com/ruckus/quickbooks-ruby/pull/484. Closing this in that case. Thanks!

Issue #515, opened 2020-04-19
 Error parsing XML: balance: undefined method `new' for BigDecimal:Class 


pjrashidi:
I've been bashing my head against the wall on this for hours. Thank you for any help.

On Ruby 2.7.0

<img width="1007" alt="image" src="https://user-images.githubusercontent.com/14811629/79684383-d0896c00-826b-11ea-923e-b141f4b24a0a.png">

My controller:
```
class OauthController < ApplicationController

  before_action :set_up_oauth_client

  def set_up_oauth_client
    oauth_params = {
      site: "https://appcenter.intuit.com/connect/oauth2",
      authorize_url: "https://appcenter.intuit.com/connect/oauth2",
      token_url: "https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer"}
      
      @oauth2_client = OAuth2::Client.new(ENV['OAUTH_CLIENT_ID'], ENV['OAUTH_CLIENT_SECRET'], oauth_params)    
    end  
    
  def authenticate
    redirect_uri = 'http://localhost:3000/oauth_callback'
    @grant_url = @oauth2_client.auth_code.authorize_url(redirect_uri: redirect_uri, response_type: "code", state: SecureRandom.hex(12), scope: "com.intuit.quickbooks.accounting")
    @quickbooks_credentials = QuickbooksCredential.first
      if @quickbooks_credentials
        qb_access_token = @quickbooks_credentials.access_token
        qb_refresh_token = @quickbooks_credentials.refresh_token
        access_token = OAuth2::AccessToken.new(@oauth2_client, qb_access_token, refresh_token: qb_refresh_token)
        
        Quickbooks.sandbox_mode = true
        
        service = Quickbooks::Service::Customer.new
        service.company_id = @quickbooks_credentials.realm_id
        service.access_token = access_token
        customers = service.all
      end
  end

  def callback
    if params[:state].present?
      redirect_uri = 'http://localhost:3000/oauth_callback'
      if resp = @oauth2_client.auth_code.get_token(params[:code], redirect_uri: redirect_uri)
        QuickbooksCredential.first.update_attributes(
          access_token: resp.token, 
          access_token_expires_at: Time.at(resp.expires_at), 
          refresh_token: resp.refresh_token,
          refresh_token_expires_at: resp.params["x_refresh_token_expires_in"].seconds.from_now,
          realm_id: params[:realmId])
        redirect_to oauth_url
      end
    end
  end

end

```

my full trace:

[...]

PR #516, opened 2020-04-20
Update Readme about roxml BigDecimal deprecation

pjrashidi:
BigDecimal deprecation in ruby 2.7.0 breaks roxml dependency on this gem

https://github.com/ruckus/quickbooks-ruby/issues/515
https://github.com/Empact/roxml/issues/64


ruckus:
Thank you. Looks like we're waiting on that Empact PR to get merged.

PR #518, opened 2020-05-08
Global minorversion

ruckus:


lib/quickbooks/service/preferences.rb
```diff
@@ -3,8 +3,7 @@ module Service
     class Preferences < BaseService
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
-        url = super(query, start_position, max_results, options)
-        "#{url}&minorversion=#{Quickbooks::Model::Preferences::MINORVERSION}"
+        super(query, start_position, max_results, options)
       end
```
drewish, line 7: Does this method need to be here? Seems like it could be removed and just let super hand it.

lib/quickbooks/service/item.rb
```diff
@@ -8,18 +8,16 @@ def delete(item)
...
       def fetch_by_id(id, params = {})
-        url = "#{url_for_base}/item/#{id}?minorversion=#{Quickbooks::Model::Item::MINORVERSION}"
+        url = "#{url_for_base}/item/#{id}"
         fetch_object(model, url, params)
       end
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
-        url = super(query, start_position, max_results, options)
-        "#{url}&minorversion=#{Quickbooks::Model::Item::MINORVERSION}"
+        super(query, start_position, max_results, options)
```
drewish, line 20: Seems like this and `url_for_resource` could be dropped in favor of the parent class's implementation now.

lib/quickbooks/service/account.rb
```diff
@@ -9,7 +9,7 @@ def delete(account)
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
         url = super(query, start_position, max_results, options)
-        "#{url}&minorversion=#{Quickbooks::Model::Account::MINORVERSION}"
+        url
```
drewish, line 12: This method could go right?

lib/quickbooks/service/preferences.rb
ruckus: @drewish you're totally right, these are unnecessary. Thanks!

lib/quickbooks/service/customer.rb
```diff
@@ -9,17 +9,16 @@ def delete(customer)
 
       def url_for_resource(resource)
         url = super(resource)
-        "#{url}?minorversion=#{Quickbooks::Model::Customer::MINORVERSION}"
+        url
       end
```
drewish, line 13: this one seems like it could go too right?

lib/quickbooks/service/invoice.rb
```diff
@@ -7,17 +7,16 @@ def delete(invoice)
       end
 
       def url_for_resource(resource)
-        url = super(resource)
+        super(resource)
```
drewish, line 10: This too?

lib/quickbooks/service/invoice.rb
```diff
@@ -7,17 +7,16 @@ def delete(invoice)
...
       def fetch_by_id(id, params = {})
-        url = "#{url_for_base}/invoice/#{id}?minorversion=#{Quickbooks::Model::Invoice::MINORVERSION}"
+        url = "#{url_for_base}/invoice/#{id}"
         fetch_object(model, url, params)
       end
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
-        url = super(query, start_position, max_results, options)
-        "#{url}&minorversion=#{Quickbooks::Model::Invoice::MINORVERSION}"
+        super(query, start_position, max_results, options)
```
drewish, line 19: and this?

lib/quickbooks/service/customer.rb
```diff
@@ -9,17 +9,16 @@ def delete(customer)
...
       def fetch_by_id(id, params = {})
-        url = "#{url_for_base}/customer/#{id}?minorversion=#{Quickbooks::Model::Customer::MINORVERSION}"
+        url = "#{url_for_base}/customer/#{id}"
         fetch_object(model, url, params)
       end
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
-        url = super(query, start_position, max_results, options)
-        "#{url}&minorversion=#{Quickbooks::Model::Customer::MINORVERSION}"
+        super(query, start_position, max_results, options)
```
drewish, line 21: this too?

drewish:
Do we have to worry about setting the global version introducing breaking changes? Like if I'm upgrading do I want to lock this at a lower version?

ruckus:
@drewish 

> Do we have to worry about setting the global version introducing breaking changes? Like if I'm upgrading do I want to lock this at a lower version?

Good question, I wondered about that too. From what I can tell QBO uses minorversions on data reads, potentially returning more info than before. I have yet to see if alter the format / payload depending on a version change. In other words, bumping the minorversion tells QBO to return additional data than before (e.g. `Customer#customer_type_ref`). I dont think its used on writes.

Thus, I __think__ it should be safe to use and not introduce backwards incompatibilities?

drewish:
That makes sense. It kind of an interesting versioning strategy if only adds additions. Typically versions are used to introduce breaking changes... so glad they haven't had to do that I guess?

Issue #520, opened 2020-05-26
gsub/NoMethodError Issue with queries using numbers


jethrodaniel:
If you pass a number to the query builder, you'll get a no method error.

For example, this code will blow up with a `NoMethodError`.

```
util = Quickbooks::Util::QueryBuilder.new
clause = util.clause('Id', '=', 4)
```
https://github.com/ruckus/quickbooks-ruby/blob/654effc06366dcb07e2a142cff3303b7a6e06912/lib/quickbooks/util/query_builder.rb#L12-L25

We should be calling `#to_s` here to ensure the `gsub` doesn't fail.

ruckus:
Good call. Thanks for the heads up

jethrodaniel:
adding a pr and test, one sec

PR #521, opened 2020-05-26
ensure we `#to_s` any non-date, non-array query values

jethrodaniel:
Before this was applied, the following code would blow up with a
`NoMethodError`.

```
util = Quickbooks::Util::QueryBuilder.new
clause = util.clause('Id', '=', 4)
```

fixes https://github.com/ruckus/quickbooks-ruby/issues/520


ruckus:
Awesome, your PR is way better. Thanks!

jethrodaniel:
Thanks @ruckus, never had a maintainer respond so quickly. 👍

PR #527, opened 2020-07-10
Fix #526, Quickbooks::PurchaseOrder::MINORVERSION reference

### LINKED ISSUE #526: uninitialized constant Quickbooks::Model::PurchaseOrder::MINORVERSION
PurchaseOrder appears to be using a different MINORVERSION which is not defined. 

I am receiving this error when attempting to query PurchaseOrders through the service  on `1.0.11`.

```ruby
po = order_service.find_by(:doc_number, "test").entries.first
```

vanboom:
PurchaseOrder was referencing a class level const that has been removed.  I did not look deeply enough yet to determine how this error made it through the test suite.

### DIFF
--- a/lib/quickbooks/service/purchase_order.rb
+++ b/lib/quickbooks/service/purchase_order.rb
@@ -7,13 +7,13 @@ module Quickbooks
       end
 
       def fetch_by_id(id, params = {})
-        url = "#{url_for_base}/purchaseorder/#{id}?minorversion=#{Quickbooks::Model::PurchaseOrder::MINORVERSION}"
+        url = "#{url_for_base}/purchaseorder/#{id}?minorversion=#{Quickbooks.minorversion}"
         fetch_object(model, url, params)
       end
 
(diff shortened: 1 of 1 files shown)


lib/quickbooks/service/purchase_order.rb
```diff
@@ -7,13 +7,13 @@ def delete(purchase_order)
       end
 
       def fetch_by_id(id, params = {})
-        url = "#{url_for_base}/purchaseorder/#{id}?minorversion=#{Quickbooks::Model::PurchaseOrder::MINORVERSION}"
+        url = "#{url_for_base}/purchaseorder/#{id}?minorversion=#{Quickbooks.minorversion}"
         fetch_object(model, url, params)
       end
 
       def url_for_query(query = nil, start_position = 1, max_results = 20, options = {})
```
vanboom, line 14: Suggest adding a test for url_for_query since it is overridden here.

ruckus:
Thank you! Yes, good question about how it passed the tests. Searching the codebase for `MINOR` shows that was the last usage. I will cut a new gem version

PR #530, opened 2020-08-09
Bugfix: Do not match percent sign when parsing Report rows

atemena:
## Issue
Getting an `invalid value for BigDecimal()` error when trying to parse percentages like "-51.71 %".

## Change
Fix third alternative in regex to not match negative numbers with trailing non digit characters


ruckus:
Thank you!

PR #536, opened 2020-12-02
added ship from address reference to invoice

seutsey:
I have added a reference in the Invoice model for ShipFromAddr (ship_from_address).  This was missing and something my organization needed to access


ruckus:
@seutsey Thank you!

PR #537, opened 2021-01-05
Purchase CDC Endpoint

nathan-mots:
This allows usage of the Change Data Capture Endpoint with Purchases as well as updating the README to account for this.


ruckus:
@nathan-mots Thank you! this rocks

PR #538, opened 2021-01-11
bump version to 1.0.15

somayehbeladi:
Updated the version to 1.0.15 to Add
- Purchase CDC endpoint
- Ship from address reference to invoice


ruckus:
Thanks for the prompt to grease the wheels! Just pushed `1.0.15` to RubyGems

PR #543, opened 2021-02-08
fix: Reject strings with newlines as being numeric.

bcarr092:
We ran into a production case where an invoice number contained a newline character. The current method of detecting if the string that contained the newline was returning that the string was in fact numeric when it wasn't. This caused the call to `BigNumber` to raise an exception.

The proposed fix is to try to parse the value as a BigNumber and if it false use the string value.

### DIFF
--- a/lib/quickbooks/model/report.rb
+++ b/lib/quickbooks/model/report.rb
@@ -38,9 +38,13 @@ module Quickbooks
           value = el.attr('value')
 
           next nil if value.blank?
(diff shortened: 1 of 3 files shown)


ruckus:
@bcarr092 Thank you! Makes sense. 

I wonder if doing the double `BigDecimal` parsing could be saved by a single parse and a begin/rescue block. But that could also be a premature optimization. I am going to guess its fine in this scenario.

Thanks again.

PR #544, opened 2021-02-08
chore: Cleanup code in fix.

bcarr092:
Updating fix based on the feedback provided [here](https://github.com/ruckus/quickbooks-ruby/pull/543#issuecomment-775351818).

Thanks for the feedback @ruckus!


ruckus:
Thank you!

PR #550, opened 2021-04-27
Allow Faraday adapter used in quickbooks-ruby to be configurable 

danielnho22:
Currently, `NetHttp` is used by default for Faraday. If `quickbooks-ruby` wants to be used in a thread-safe context, we should replace the `net-http` adapter with something like `net-http-persistent`.

This allows that to be done by setting `Quickbooks.http_adapter`

### DIFF
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -194,6 +194,7 @@ module Quickbooks
   @@sandbox_mode = false
   @@logger = nil
   @@minorversion = 47
+  @@http_adapter = :net_http
 
   class << self
     def sandbox_mode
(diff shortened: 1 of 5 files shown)


ruckus:
@danielnho22 Good call. Any reason why this shouldn't just be made the default?

danielnho22:
@ruckus That's a good question. I can't find any info of why `net-http-persistent` isn't just the default in Faraday. If we're cool with it, I can make it the default for `quickbooks-ruby`.

danielnho22:
@ruckus The default is now `net-http-persistent`

danielnho22:
I imagine that most people using this gem are consistently making requests to the same servers and re-using connections to them in a thread-safe context would be preferable.

ruckus:
@danielnho22 thanks!!!

PR #552, opened 2021-04-28
Change default Faraday adapter to :net-http

danielnho22:
In response to https://github.com/ruckus/quickbooks-ruby/pull/550 

We added the ability to configure the adapter for Faraday, but also changed the default to `net-http-persistent`.

`Net::HTTP::Persistent` is an optimization that is supposedly to be thread-safe, but I have not personally tested it enough to make it the default adapter. I don't want to mess with other integrations.

While I test this a bit more, I think we should move this back to use `net-http`


ruckus:
@danielnho22 fair enough, better to be safe.

ruckus:
@danielnho22 Thanks again. I just released `1.0.17` with your PR / feature.

Nitrodist:
We're definitely seeing errors in a threaded environment with net-http opted in :)

ruckus:
@Nitrodist have you tried switching the adapter to `net-http-persistent`? supposedly its thread-safe but my environment is not threaded so I dont have any direct experience with it.

PR #554, opened 2021-07-06
Fixes #439 - Don't quote boolean values in queries

### LINKED ISSUE #439: IntuitRequestException thrown when querying using boolean value
When using the [`Service#find_by`](https://github.com/ruckus/quickbooks-ruby/blob/4d788fbc9bf49cd849199b12a667dc3cce4c20cf/lib/quickbooks/service/service_crud.rb#L28-L34) method, it appears that passing a boolean value results in `IntuitRequestException` being thrown.  I'm pretty sure this is due to the single quotes found in the string interpolation.

Throws:

```ruby
Quickbooks::Service::Employee.new({company_id: "...", access_token: "..."}).find_by(:active, true)
# => Quickbooks::IntuitRequestException: An application error has occurred while processing your request: System Failure Error: java.lang.String cannot be cast to java.lang.Boolean
```

Throws (note the single quotes):

```ruby
Quickbooks::Service::Employee.new({company_id: "...", access_token: "..."}).query("select * from employee where Active = 'true'")
# => Quickbooks::IntuitRequestException: An application error has occurred while processing your request: System Failure Error: java.lang.String cannot be cast to java.lang.Boolean
```

Works (note the lack of single quotes):

```ruby
Quickbooks::Service::Employee.new({company_id: "...", access_token: "..."}).query("select * from employee where active = true")
```

jaredmoody:
- Update Quickbooks::Util::QueryBuilder to handle boolean values
- Update Quickbooks::Service::ServiceCrud#find_by to use the QueryBuilder so that booleans are properly handled

### DIFF
--- a/lib/quickbooks/service/service_crud.rb
+++ b/lib/quickbooks/service/service_crud.rb
@@ -33,7 +33,10 @@ module Quickbooks
         if field.class == Symbol
           field = field.to_s.camelcase
         end
-        q = "select * from %s where %s = '%s'" % [model.resource_for_singular, field, selector]
+
+        clause = Quickbooks::Util::QueryBuilder.new.clause(field, '=', selector)
+        q = "select * from %s where %s" % [model.resource_for_singular, clause]
+
         self.query(q, options)
       end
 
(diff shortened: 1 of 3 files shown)


ruckus:
@jaredmoody thank you!!

PR #556, opened 2021-08-12
Added linked_transactions to bill_line_item

ihor-wrl:
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/most-commonly-used/bill  
'Zero or more transactions linked to this Bill object. The LinkedTxn.TxnType can be set to PurchaseOrder, BillPaymentCheck or if using Minor Version 55 and above ReimburseCharge. Use LinkedTxn.TxnId as the ID of the transaction.'

### DIFF
--- a/lib/quickbooks/model/bill_line_item.rb
+++ b/lib/quickbooks/model/bill_line_item.rb
@@ -12,6 +12,8 @@ module Quickbooks
       xml_accessor :amount, :from => 'Amount', :as => BigDecimal, :to_xml => to_xml_big_decimal
       xml_accessor :detail_type, :from => 'DetailType'
 
+      xml_accessor :linked_transactions, :from => 'LinkedTxn', :as => [LinkedTransaction]
+
       #== Various detail types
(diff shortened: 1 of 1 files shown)


ruckus:
Thank you!

ihor-wrl:
> Thank you!

I gave not exact description recently. Here it is:
'Zero or more transactions linked to this object. The LinkedTxn.TxnType can be set to PurchaseOrder or if using Minor Version 55 and above it can also be set to ReimburseCharge. Use LinkedTxn.TxnId as the ID of the transaction. When updating an existing Bill to link to a PurchaseOrder a new Line must be created. This behavior matches the QuickBooks UI as it does not allow the linking of an existing line, but rather a new line must be added to link the PurchaseOrder. Over the API this is achieved by simply updating the Bill Line.Id to something new. This will ensure old bill line is deleted and the new line is linked to the PurchaseOrder. Please be aware that for this PurchaseOrder-Bill linking to work all LinkedTxn child attributes are required. See child attributes below.' 
So, to link Bill to PO, we pass Bill.linked_transactions and BillLineItem.linked_transactions.

PR #566, opened 2021-11-30
Add total_tax_specified to TransactionTaxDetailModel

danielnho22:
When creating or updating a QBO invoice and not specifying the `TotalTax`, we need to pass a field `TotalTaxSpecified` as false to QBO.

In total, this PR holds makes the following changes:

1. Allows us to set `TotalTaxSpecified` on the `TransactionTaxDetail`
2. Adds some specs for the above
3. Refactors some old specs for that model.

**NOTE**: We're using this internally and this is just so others are able to set it. Also, the reason this came up is described in this thread: https://help.developer.intuit.com/s/question/0D74R000004nqSK/detail?s1oid=00DG0000000COk8&s1nid=0DBG0000000blK7&emkind=chatterCommentNotification&s1uid=0054R00000B4S02&emtm=1638296091795&fromEmail=1&s1ext=0

### DIFF
--- a/lib/quickbooks/model/transaction_tax_detail.rb
+++ b/lib/quickbooks/model/transaction_tax_detail.rb
@@ -1,9 +1,9 @@
 module Quickbooks
   module Model
     class TransactionTaxDetail < BaseModel
-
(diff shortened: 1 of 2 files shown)


ruckus:
@danielnho22 Thank you!

PR #573, opened 2022-03-17
Add condense logs option

andymond:
Add an option to only call the logger once per request & once per response when talking to QBO to make searching logs in general & especially 3rd party log managers easier.


ruckus:
Thank you!

PR #576, opened 2022-04-14
Add Linked Transactions Array to VendorCredit

veronicaday:
Adds ability to access linked transactions on the vendor credit object. For example, if the credit was applied to a bill as payment.


ruckus:
Thank you @veronicaday

PR #586, opened 2022-11-10
Upgrade faraday and oauh2 to v2

marcrohloff:
* Use Faraday 2.0 gem
* Replace custom Faraday GZip implementation with `faraday-gzip` gem
* Relax gemspec to allow OAuth2 v2.x gem
* No longer support Ruby 2.5
* Remove extraneous `pp` in specs

Credit goes to @Zajn 
Original [PR https://github.com/ruckus/quickbooks-ruby/pull/58](https://github.com/ruckus/quickbooks-ruby/pull/578)


ruckus:
Thanks everyone.

So ... this is now the start of a `1+` and `2+` version era. I merged this PR into the new `2-stable` branch and released version `2.0` to RubyGems.

Users on Ruby 2.6+ should use 2+ of the gem.

any `1.x` versions are legacy and only people still on Ruby 2.5.8 (like me) can only use <= 1.x

Issue #588, opened 2022-11-21
TaxInclusiveAmt has inconsistent serialization/deserialization types


brianhenryhf:
Models that include `tax_inclusive_amount` use `#to_f` when serializing to XML, landing as floats, but specify no `:as` type, and default to Strings when deserialized.

Example XML string from serialization of a SalesReceipt (note `<TaxInclusiveAmt>0.0</TaxInclusiveAmt>`):
```
[...]

PR #589, opened 2022-11-22
deser tax_inclusive_amt to numeric type

brianhenryhf:
Fix for https://github.com/ruckus/quickbooks-ruby/issues/588:  match the numeric serialization type for this field, across models where used.


ruckus:
Thank you!

PR #592, opened 2022-11-25
Add support for sending remaining options with query

danielwestendorf:
+ Append all non-standard options to the query url
+ Rely on Faraday's query string encoding

Some endpoints (like invoice) require additional parameters on the query URL to get additional attributes (like include=invoiceLink).


vanboom:
:+1:

ruckus:
Thank you!

Issue #587, opened 2022-11-18
Service::AccessToken#disconnect causes ArgumentError (wrong number of arguments (given 4, expected 3))


JeremyC-za:
Hey, as the title says, calling `Service::AccessToken#disconnect` results in an exception. 

This **only happens in in the 2.0 release**, it works as expected in 1.0.21. The error seems to come from Faraday, which has me looking at [this recent commit](https://github.com/ruckus/quickbooks-ruby/commit/32e892bde6edab3fb87513e15ee7928faed02da4).

Here's the full trace:
```
faraday (1.10.2) lib/faraday/request/authorization.rb:48:in `initialize'
faraday (1.10.2) lib/faraday/dependency_loader.rb:21:in `new'
faraday (1.10.2) lib/faraday/dependency_loader.rb:21:in `new'
faraday (1.10.2) lib/faraday/rack_builder.rb:57:in `build'
faraday (1.10.2) lib/faraday/rack_builder.rb:175:in `block in to_app'
faraday (1.10.2) lib/faraday/rack_builder.rb:174:in `each'
faraday (1.10.2) lib/faraday/rack_builder.rb:174:in `inject'
faraday (1.10.2) lib/faraday/rack_builder.rb:174:in `to_app'
faraday (1.10.2) lib/faraday/rack_builder.rb:167:in `app'
faraday (1.10.2) lib/faraday/rack_builder.rb:154:in `build_response'
faraday (1.10.2) lib/faraday/connection.rb:516:in `run_request'
faraday (1.10.2) lib/faraday/connection.rb:283:in `post'
quickbooks-ruby (2.0) lib/quickbooks/service/access_token.rb:29:in `disconnect'
```

Lauren-penner:
Faraday is not being forced above 1.x. If you are not ready to upgrade to Faraday 2.x you could overwrite the `disconnect` method to use the old authentication helper method `basic_auth`.
`f.basic_auth(oauth.client.id, oauth.client.secret)`

jaredmoody:
Monkey patching the gem is a workaround, but not a solution. If the gem isn't going to require Faraday 2.x, then it should handle using both instead of erroring when using an allowed dependency.

ashkulz:
@JeremyC-za this has been fixed in the 2.0.1 release :+1:

PR #593, opened 2022-12-20
Faraday Versions

jaredmoody:
Both Faraday 1.x and 2x are allowed dependencies, but the released code currently breaks with Faraday 1.x - one example is reported in #587.

If 2.x is not going to be required, the gem needs to be tested against and support both versions.

This PR adds Faraday 1.x to the test matrix, with the goal of revealing broken tests so they can be addressed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,11 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["2.6", "2.7", "3.0", "3.1", ruby-head]
+        env:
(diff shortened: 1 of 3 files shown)


jaredmoody:
Disconnecting the Access token was the only failing test. The failures on ruby-head seem unrelated to this PR.

ashkulz (approved):
This looks good, @jaredmoody! Any chance this could be reviewed, @ruckus?

ruckus:
Thanks @jaredmoody and @ashkulz . I just released `2.0.1` with this PR.

PR #596, opened 2023-04-13
Use Faraday::UploadIO instead of UploadIO

ngan:
This changes all reference of the deprecated `::UploadIO` for `Faraday::UploadIO`. `multipart-post 2.2.0` deprecates top-level `UploadIO`: https://github.com/socketry/multipart-post/commit/cb18a32ed9522af9e4072195d0411a8281c6a061#diff-f192e4f2c6b7ce852ab54997f6dec1141f625ccd9d02a002b1a8adc39033c250

I decided to just use Faraday's alias of `UploadIO` since they check the first of `mutlipart-post` you're on and alias the constant properly. I also added `>= 1.0.4` for `multipart-post`...since `Faraday::UploadIO` wasn't introduced until 1.0.2 anyways. So, I'm not sure how this was working for folks with `multipart-post 1.0.0 or 1.0.1`: https://github.com/ruckus/quickbooks-ruby/blob/2.0.2/lib/quickbooks/service/upload.rb#L23

### DIFF
--- a/lib/quickbooks/service/base_service.rb
+++ b/lib/quickbooks/service/base_service.rb
@@ -224,7 +224,7 @@ module Quickbooks
         if metadata
           standalone_prefix = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
           meta_data_xml = "#{standalone_prefix}\n#{metadata.to_xml_ns.to_s}"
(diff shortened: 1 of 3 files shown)


ngan:
@ruckus mind taking a look?

PR #600, opened 2023-05-31
Fintechqb 1967 enable service level logging

andymond:
- Update log check to be configurable by whatever class includes the logging module, preserving default functionality.
- Support the case where we would like to have some services log without enabling logging through our application instance.


ruckus:
Thank you!

PR #608, opened 2024-04-09
Add element to IntuitRequestException and message

dmorehouse:
Have you error received a field error like "String length specified does not match the supported length" and wondered which of the many fields caused the issue? Me too, this PR changes all that :-)

My proposal makes the element (field) available in the IntuitRequestException for programatic access to the element name associated with the error message. It also adds the element name if present (it's optional) to the exception message so that messages are prepended with the specific field if present.

Tradeoffs: Adding the element name to the exception message could break existing code if logic was based on the EXACT message returned. Logic based on an exact match instead of an includes would be brittle to message changes from QBO and is a bad practice. I think it's a good tradeoff to add the element name to the message which is generally used everywhere as by default error messages will now indicate which element they apply to. The message already concatenates the underlying error message and error detail fields and this seemed like very useful context to add to the message by default.

### DIFF
--- a/lib/quickbooks-ruby.rb
+++ b/lib/quickbooks-ruby.rb
@@ -275,7 +275,7 @@ module Quickbooks
   class UnsupportedOperation < Error; end
 
   class IntuitRequestException < Error
-    attr_accessor :message, :code, :detail, :type, :intuit_tid, :request_xml, :request_json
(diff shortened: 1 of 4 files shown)


dmorehouse:
I'm recreating this against the `2-stable` branch

PR #613, opened 2024-06-26
Loosen faraday-gzip to a version that can use > zlib-2.1.1

technicalpickles:
I was looking to update our app from using zlib 2.1.1 to the latest zlib-3.1.1, but bundler wasn't able to update it. I tracked down the dependency chain to:

- faraday-gzip-1.0.1 depends on  zlib `~> 2.1`, which resolves to zlib-2.1.1
- quickbook-ruby depends on faraday-gzip `~> 1.0` which resolves to faraday-gzip-1.0.1

This PR tries to loosen the dependency to `>= 1.0`. I'm open to other variations that would allow faraday-gzip-2 though.

### DIFF
--- a/quickbooks-ruby.gemspec
+++ b/quickbooks-ruby.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency 'multipart-post' # promiscuous mode
   gem.add_dependency 'faraday', '< 3.0'
   gem.add_dependency 'faraday-multipart', '~> 1.0', '>= 1.0.4'
-  gem.add_dependency 'faraday-gzip', '~> 1.0'
+  gem.add_dependency 'faraday-gzip', '>= 1.0'
 
   gem.add_development_dependency 'rake'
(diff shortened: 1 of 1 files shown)


kevGalway:
We have the same problem so eager for solution to this too. 

We are on latest version of ruby (3.3.3) meaning zlib 3.1.1 is a standard/default gem (https://stdgems.org/zlib/)

to update faraday to 2.x we need to update quickbooks-ruby to 2.x
quickbooks-ruby2.x depends on faraday-gzip (~> 0.1) to faraday-gzip (~> 1.0)
faraday-gzip before 2.0 depends on zlib 2.x and only starting to use zlib 3.x from version 2.0.0

@technicalpickles purposed solutions should work for us.

ruckus:
Thank you @technicalpickles . Makes sense.

technicalpickles:
@ruckus Thank you! Any chance of getting a release with this out as well?

ruckus:
Hi @technicalpickles I just released `2.0.4` with your PR. Thanks again!!

PR #616, opened 2024-11-14
Add ship from address reference to sales receipt

rchekaluk:
Analogous to https://github.com/ruckus/quickbooks-ruby/pull/536 for Sales Receipt.

Also added some minor additional expects for coverage.


rchekaluk:
Ping @ruckus this seems like a small but reasonable addition to coverage of the Quickbooks API.

PR #617, opened 2024-11-19
added pdf method to refund_receipt.rb

nicho1991:
Refund receipt is missing pdf method, the api does support it, so lets add it :)

https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/refundreceipt#get-a-refund-receipt-as-pdf


ruckus (approved):

ruckus:
Thank you! @nicho1991

Issue #622, opened 2025-01-22
Quickbooks deprecating support for minor versions <75 of the Online Accounting API on August 1, 2025


EvanGrote:
I received an email this morning from the Intuit Developer Group that beginning August 1, 2025, they will be discontinuing support for minor versions 1 through 74 of the QuickBooks Online Accounting API.

[Here is the blog linked in the email announcing the change.](https://blogs.intuit.com/2025/01/21/changes-to-our-accounting-api-that-may-impact-your-application/)

From the email:
> In your API request, if you specify a value for the minorversion parameter that is less than 75, it will be ignored and the system will respond with data corresponding to minor version 75.
If you don't specify a minor version in your request, the response will default to minor version 75 rather than the base version.

Currently, [this gem uses minor version 47](https://github.com/ruckus/quickbooks-ruby/blob/568b0aa473a3d56858b400f2858fbfd7c0c32078/lib/quickbooks-ruby.rb#L196), so the minor version will need to be updated to avoid breaking changes when Intuit begins responding to version 47 requests with version 75 responses.

bretweinraub:
I just came in to post the same thing.....

nicho1991:
Do we know if there will be any incombatility on this upgrade? 47 to 75 is afterall, a pretty big jump!

nehresma:
> Do we know if there will be any incombatility on this upgrade? 47 to 75 is afterall, a pretty big jump! also, noting the blog mentions that using any sub 75 minorversion, it will just reponse as if it was 75, so its probably important to do some testing on the 75 version before July

@nicho1991 I believe it will depend on how your application is using the API.  I'm in the process of reading through the release notes for all the minor versions to see how I will be impacted:  https://developer.intuit.com/app/developer/qbo/docs/learn/explore-the-quickbooks-online-api/minor-versions#minor-version-summary

ruckus:
From my playing around with various versions in the past it just unlocks different fields. The shape of the data is the same. But sometimes new fields are exposed.

So it should be backwards compatible with any current client code you might have. 

That has been my experience at least.

nehresma:
@ruckus Yes, I agree.  FWIW, I just finished the process of upgrading all of our code to use minor version 75 and for our purposes, it was completely backwards compatible.  We use:

* Account
* Class
* CompanyCurrency
* Customer
* Deposit
* Invoice
* Payment
* Preferences
* RefundReceipt
* TaxCode
* TaxRate

[...]

PR #623, opened 2025-01-30
feat: set minor version to 75 & add docs

thestelz:
* Set the minor version to 75 as that will be the new default via QBO API

* Add a section via the readme about the minor version and why the default is 75

closes ruckus/quickbooks-ruby#622


AlexYakovlevCa:
Hey @thestelz  , are you sure this change doesn't affect anything? I know they're enforcing this.

